PR #18, opened 2023-11-03
Working Blinky with MCAL

samparent97:


clarawong20 (approved):
Great work on getting it to work with the board!

PR #21, opened 2023-11-10
User/freerb/convert to handles

BlakeFreer:


firmware/TMS/mcal/stm32f767/.mxproject
```diff
@@ -0,0 +1,19 @@
+[PreviousLibFiles]
```
samparent97, line 1: Is this something we want to have pushed to git?

firmware/TMS/mcal/stm32f767/Makefile
```diff
@@ -0,0 +1,196 @@
+##########################################################################################################################
+# File automatically-generated by tool: [projectgenerator] version: [3.18.0-B7] date: [Thu Nov 09 20:56:00 EST 2023]
```
samparent97, line 2: Is the auto gen makefile used?

firmware/shared/util/peripheral.h
```diff
@@ -0,0 +1,17 @@
...
+/// @date 2023-11-08
+
+#ifndef SHARED_UTIL_PERIPHERAL_H_
+#define SHARED_UTIL_PERIPHERAL_H_
+
+#include "shared/util/noncopyable.h"
+
+namespace shared::util {
+
+class Peripheral {//: private Noncopyable {
```
samparent97, line 11: Might be missing it, but what is the purpose of this

BlakeFreer: It's generic properties (like move semantics) that are inherited by all peripherals. Currently it is empty but I want to keep it available for the future. Since we don't reference anything bu the parent Peripheral class, this does not break the "avoiding abstraction" goal

firmware/TMS/mcal/stm32f767/Makefile
BlakeFreer: No it is not

firmware/TMS/mcal/stm32f767/.mxproject
BlakeFreer: It's generated from the ioc, we should decide on Thursday how much to commit and how much to generate locally

samparent97 (approved):

PR #41, opened 2024-01-26
Generate CubeMX project from CMake

### LINKED ISSUE #31: CubeMX Command code generation
Currently, the developer must manually open CubeMX and generate code for each .ioc file.

This is slow and can be done from the command line.

Main task:
Figure out how to use the command line to generate code from an ioc

Once this is done:
Edit the demo project mcal/stm/cubemx CMake file to call this command automatically. It should only generate new code if the ioc has been modified.

### LINKED ISSUE #32: Custom CubeMX makefile targets
Currently, the cubemx CMake builds the "all" target of the auto-generated Makefile, which builds all object files but then crashes when generating the .elf file.

This works, since the CMake only requires the object files from Makefile, and `make` crashes after these objects are compiled, and the crash does not halt the rest of cmake.

We should try to find a cleaner way to build the cubemx files. Someone should look into custom code generation through CubeMX and see if there is a way to modify the generated Makefile. If so, simply adding the target
`objects: $(OBJECTS)`
would allow us to call `make objects` and properly generate all of the object files without crashing.

Micnasr:
Closes #31 

Also closes #32 by modifying the generated Makefile after generating it.


firmware/projects/DemoProject/platforms/stm32f767/cubemx/CMakeLists.txt
```diff
@@ -3,7 +3,22 @@
 
 message(STATUS "Building CubeMX objects with the auto-generated Makefile")
 if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/Makefile")
-	message(FATAL_ERROR "No auto-generated Makefile was found. Open the .ioc file in STM32CubeMX and click [Generate Code]")
+	find_program(STM32CubeMX_PATH STM32CubeMX.exe)
+
+	if(NOT STM32CubeMX_PATH)
+		message(FATAL_ERROR "STM32CubeMX.exe not found in PATH")
```
samparent97, line 9:
```suggestion
		message(FATAL_ERROR "STM32CubeMX.exe not found in PATH. Open the .ioc file in STM32CubeMX and click [Generate Code]")
```

samparent97 (commented):
Nice! Just fix up the message if `STM32CubeMX.exe` is not found.

BlakeFreer:
Nothing is generated for me. I believe the issue is that I don't have java installed. Can you add instructions in `firmware/README` regarding what java version is required and how to configure it (i.e. does it need to be on path)?

firmware/projects/DemoProject/platforms/stm32f767/cubemx/CMakeLists.txt
BlakeFreer: I don't think this will work,. CMake will report FATAL_ERROR when it reaches the stage where it looks for CubeMX in the path, regardless of whether the user has manually clicked [generate code].

Since the user needs CubeMX installed for manual and automatic generation, I suggest we enforce it to be on the path

samparent97: Ahh good point. The only issue is that Unix based devices will not have the `.exe` extension.

samparent97: If you manually generate, the make file will exist and it will not attempt to generate. This might be a problem however. We should maybe add a condition that the `board_config.ioc` file has not changed

Micnasr: > If you manually generate, the makefile will exist and it will not attempt to generate. This might be a problem however. We should maybe add a condition that the `board_config.ioc` file has not changed

Does this mean that it should only generate if the .ioc file has changed? Also regarding the comment above, should I implement the suggestion or keep it as is to enforce it to be on path?

BlakeFreer: Yes it should only regen if the ioc changes. Otherwise the build process will be very long for every build.
If we were using Make instead of CMake, then the recipe
```
<any cubemx generated file>: board_config.ioc
    <Java generate call>
```
Would regenerate only when the build_config.ioc is updated after the cubemx file is generated. 

I think you can do a similar thing with CMake

BlakeFreer: And it should require STM32cubemx on path

Micnasr: The only way I can think of doing it is storing the timestamp of the last build and reading it once we try to build again and seeing if it matches. Would that be a good solution or is there a better way to do this?

samparent97: I think the best thing to do at this point is to get this working on a Unix machine. The team raspi would be nice and it is always accessible. CMake allows you to check if you are on a Windows OS or a Unix based OS so if there are major differences you could branch on this condition.

BlakeFreer: > The only way I can think of doing it is storing the timestamp of the last build and reading it once we try to build again and seeing if it matches. Would that be a good solution or is there a better way to do this?

I'm pretty sure CMake has this functionality built in

Micnasr: > I think the best thing to do at this point is to get this working on a Unix machine. The team raspi would be nice and it is always accessible. CMake allows you to check if you are on a Windows OS or a Unix based OS so if there are major differences you could branch on this condition.

how would the command differ if we are trying to run it in Unix? How can I even test that, I do not own any non windows machines?

BlakeFreer: resolved by my modifications

BlakeFreer (dismissed):
approved

BlakeFreer (approved):

samparent97 (approved):

PR #45, opened 2024-02-04
converts peripherals and app objects to use abstract classes

BlakeFreer:


firmware/shared/periph/gpio.h
```diff
@@ -4,27 +4,22 @@
...
-	{ obj.SetHigh() } -> std::same_as<void>;
-	{ obj.SetLow() } -> std::same_as<void>;
-	{ obj.Set(value) } -> std::same_as<void>;
-
-	std::is_base_of_v<util::Peripheral, T>;
+class DigitalOutput : public util::Peripheral {
+public:
+    virtual void SetHigh() = 0;
+    virtual void SetLow() = 0;
+    virtual void Set(bool value) = 0;
```
samparent97, line 20: We should add a toggle method.
```suggestion
    virtual void Toggle() = 0;
```

firmware/mcal/stm32f767/periph/adc.h
```diff
@@ -5,36 +5,36 @@
...
+
+    uint32_t Read() override {
+        /// @todo should there be a standard output range?
+        /// if so, the conversion would require knowing the resolution and
+        /// alignment settings
+
+        /// @todo change this to be non-blocking, currently just for demo
+        Start();
+        while (HAL_ADC_PollForConversion(hadc_, 1000) != HAL_OK)
+            ;
```
samparent97, line 33: Why is the `;` on a new line here? Guessing it's a clang-format thing?

samparent97 (dismissed):

firmware/mcal/stm32f767/periph/adc.h
BlakeFreer: Yes it's a clang issue that doesn't have a supported solution. That line of code is also incorrect so I will fix it altogether.

The issue is you should not repeatedly poll the adc, as this renders the 1000ms timeout useless

firmware/shared/periph/gpio.h
samparent97: Ok, we will leave it for now until there is a need.

samparent97 (approved):

PR #48, opened 2024-02-06
removes drivers folder from git stats

BlakeFreer:
See https://github.com/github-linguist/linguist/blob/master/docs/overrides.md for explanation

### DIFF
--- a/.github/.gitattributes
+++ b/.github/.gitattributes
@@ -1,0 +1,1 @@
+../firmware/drivers linguist-vendored


samparent97 (approved):
Another possibility for the large amount of C over C++ could be the header naming as .h instead of .hpp.

BlakeFreer:
> Another possibility for the large amount of C over C++ could be the header naming as .h instead of .hpp.

It's more correct just to remove the folder entirely since we didn't write it. If we relabel it as c++ headers then the balance will be skewed the other way.

Also every driver import looks like `#include "file.h"` so renaming would break every import

PR #49, opened 2024-02-07
Making cubemx/generate.mk standalone

### LINKED ISSUE #47: make cubemx/generate.mk standalone
Currently the `generate.mk` file used to generate the cubemx code from the .ioc file requires input from the adjacent CMakeLists.

In particular CMake provides the `CUBEMX_PATH` variable which specifies the location of the STM32CubeMX program, and also provides the `IOC_FILE` variable which is just `board_config.ioc`

It would be better if generate.mk does not require any input from CMake, since it could then be run standalone to generate the CubeMX code by calling `make --file generate.mk`

1. Populate the `IOC_FILE` variable in generate.mk with `board_config.ioc` and remove the variable from the CMakeLists execute_process command. This variable must be explicitly set somewhere, so it might as well be in generate.mk since this is the only place where it is required
2. Find a way to migrate from CMake's `find_program()` command so that generate.mk does not need `CUBEMX_PATH` passed in from CMake. One option is to require the developer to place the program on the path, there may be other solutions.

Apply these changes to all projects under firmware/projects that support the stm32f767 platform

Micnasr:
Closes #47 

- Added hardcoded .ioc file name in generate.mk
- Added logic to grab the STM32CubeMX path depending on platform (windows / linux)

Testing:
I was able to generate by only running the command below with no extra arguments:

`make -f generate.mk`

### DIFF
--- a/firmware/projects/DemoProject/platforms/stm32f767/cubemx/CMakeLists.txt
+++ b/firmware/projects/DemoProject/platforms/stm32f767/cubemx/CMakeLists.txt
@@ -9,7 +9,7 @@ if(NOT STM32CubeMX)
 endif()
 
 execute_process(
-	COMMAND ${CMAKE_MAKE_PROGRAM} "--file" "generate.mk" "CUBEMX_PATH=${STM32CubeMX}" "IOC_FILE=${CMAKE_CURRENT_SOURCE_DIR}/board_config.ioc"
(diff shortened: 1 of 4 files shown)


BlakeFreer (approved):
Approved on windows. I added the modifications to the TMS project

samparent97 (approved):

PR #57, opened 2024-03-08
Implemented Target Deep Clean

### LINKED ISSUE #51: Deep clean
The current build system cleans by deleting files in the `firmware/build` directory that is created by CMake.

This does not clean the files that are generated and built by cubemx, as these files are stored within the project's `platforms/stm32f767/cubemx` folder. Therefore, to fully clean, we must manually delete all of those files.

This is time consuming and error prone as it would be easy to accidentally delete a non-generated file like `generate.mk` or `CMakeLists.txt`.

Task: When `make PROJECT=<proj> PLATFORM=<plat> deepclean` is executed from the firmware Makefile, all cubemx generated and built files should be deleted, in addition executing the regular `clean` recipe.

I do not think that the regular `clean` target should delete the cubemx files, as the cubemx generation is relatively slow and does not need to occur as frequently as cleaning the CMake build files.

I can think of one option; there may be others

- Somehow pass the `clean` command down into the cubemx folder and have some script in cubemx/ handle it
    - generate.mk would be a good script to use, you could add a clean target to it. This target would delete everything that is not in the `.gitignore`

Micnasr:
Closes #51 

Implemented Deep clean which will call original clean and remove all generated files under cubemx. Unsure if that is the way you want it to be deleted since it happens at the firmware level.

It will basically delete any folder/file under the cubemx directory that is not in the KEEP_FILE string.

TESTING:
Built the project a few times and deep cleaned it to make sure it only deleted generated content.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -12,9 +12,18 @@ clean: FORCE
 	rm $(BUILD_DIR) -rf
 
 deepclean: clean
-	@echo "Deleting generated files and directories in $(CUBEMX_DIR) that are ignored by Git"
+ifeq ($(findstring stm32,$(PLATFORM)),stm32)
+	@echo "Deleting files and directories in $(CUBEMX_DIR) that are ignored by Git"
 	@cd $(CUBEMX_DIR) && \
-		git check-ignore * | xargs -I{} rm -rf "{}"
+		for file in $$(ls -R); do \
+			if git check-ignore "$$file" >/dev/null; then \
+				echo "Deleting $$file"; \
+				rm -rf "$$file"; \
+			fi; \
+		done
+else
+	@echo "Skipping deepclean for platform $(PLATFORM)"
+endif
 
 st-flash: build
 	st-flash --reset write $(BUILD_DIR)/main.bin 0x08000000
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -4,8 +4,6 @@ PROJECT =
 BUILD_DIR = build/$(PROJECT)/$(PLATFORM)
 CUBEMX_DIR := projects/$(PROJECT)/platforms/$(PLATFORM)/cubemx
 
-KEEP_FILES := .gitignore CMakeLists.txt README board_config.ioc generate.mk
-
 build: FORCE
 	cmake -B $(BUILD_DIR) -S. -G"Unix Makefiles" -DPROJECT=$(PROJECT) -DPLATFORM=$(PLATFORM) 
 	cmake --build $(BUILD_DIR)
@@ -14,18 +12,9 @@ clean: FORCE
 	rm $(BUILD_DIR) -rf
 
 deepclean: clean
-	@echo "Deleting files and directories in $(CUBEMX_DIR) except $(KEEP_FILES)"
-	@cd $(CUBEMX_DIR) && for file in *; do \
-		if ! echo "$(KEEP_FILES)" | grep -q "\\b$$file\\b"; then \
-			if [ -d "$$file" ]; then \
-				echo "Deleting directory $$file"; \
-				rm -rf "$$file"; \
-			else \
-				echo "Deleting file $$file"; \
-				rm -f "$$file"; \
-			fi; \
-		fi; \
-	done
+	@echo "Deleting generated files and directories in $(CUBEMX_DIR) that are ignored by Git"
+	@cd $(CUBEMX_DIR) && \
+		git check-ignore * | xargs -I{} rm -rf "{}"
 
 st-flash: build
 	st-flash --reset write $(BUILD_DIR)/main.bin 0x08000000
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -2,6 +2,9 @@ PLATFORM =
 PROJECT = 
 
 BUILD_DIR = build/$(PROJECT)/$(PLATFORM)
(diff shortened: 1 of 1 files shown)


firmware/Makefile
```diff
@@ -2,6 +2,9 @@ PLATFORM =
 PROJECT = 
 
 BUILD_DIR = build/$(PROJECT)/$(PLATFORM)
+CUBEMX_DIR := projects/$(PROJECT)/platforms/$(PLATFORM)/cubemx
```
BlakeFreer, line 5: This will fail if `PLATFORM != stm32f767` since other platforms do not have a `cubemx` subdirectory

firmware/Makefile
```diff
@@ -2,6 +2,9 @@ PLATFORM =
...
+
+KEEP_FILES := .gitignore CMakeLists.txt README board_config.ioc generate.mk
```
BlakeFreer, line 7: we shouldn't rely on a global setting for these files. Since the `cubemx/.gitignore` file already describes what is stored in version control, you should use its contents instead

The [`git check-ignore`](https://git-scm.com/docs/git-check-ignore) command would be excellent for this

firmware/Makefile
```diff
@@ -10,6 +13,20 @@ build: FORCE
 clean: FORCE
 	rm $(BUILD_DIR) -rf
 
+deepclean: clean
```
BlakeFreer, line 16: good use of `clean` as a prerequisite

firmware/Makefile
```diff
@@ -10,6 +13,20 @@ build: FORCE
...
+	@echo "Deleting files and directories in $(CUBEMX_DIR) except $(KEEP_FILES)"
+	@cd $(CUBEMX_DIR) && for file in *; do \
+		if ! echo "$(KEEP_FILES)" | grep -q "\\b$$file\\b"; then \
```
BlakeFreer, line 19: if you use `git check-ignore`, then you will get a list of all files which should be deleted. Just call `rm <all those files>` instead of looping

BlakeFreer (changes requested):

firmware/Makefile
Micnasr: Is the only solution using a condition to check if the project is stm vs something else? Or is there a better approach

BlakeFreer: any code compiled by CMake is placed in the build/ directory, which is why it is so easy to clean most of the project.
We are trying to clean the code generated by CubeMX, and compiled by the generated Makefile. Since CMake doesn't directly compile this code, it goes into the cubemx/build directory

For now, check that the platform starts with stm32. That will be sufficient

firmware/Makefile
```diff
@@ -14,18 +12,9 @@ clean: FORCE
...
-				rm -rf "$$file"; \
-			else \
-				echo "Deleting file $$file"; \
-				rm -f "$$file"; \
-			fi; \
-		fi; \
-	done
+	@echo "Deleting generated files and directories in $(CUBEMX_DIR) that are ignored by Git"
+	@cd $(CUBEMX_DIR) && \
+		git check-ignore * | xargs -I{} rm -rf "{}"
```
BlakeFreer, line 17: looks good, i like the use of a pipe.

BlakeFreer (changes requested):

firmware/Makefile
BlakeFreer: I just realized that since `git check-ignore *` only searches the current directory, it will not report folders that have even 1 non-ignored file.

the `find` command will recursively report all files and subdirectories, allowing `git check-ignore` to individually assess each

```console
find | xargs git check-ignore | xargs rm -rf
```

also notice that the default behaviour of `xargs` is to places the contents after the command, so you do not need the `-I{}` replacement syntax

BlakeFreer: Is there a reason for the `&& \`? IIRC this executes the two commands in sequence, but can't this be achieved by simply writing the `git check-ignore` command on the next line?

Micnasr: > Is there a reason for the `&& \`? IIRC this executes the two commands in sequence, but can't this be achieved by simply writing the `git check-ignore` command on the next line?

Yeah its the same thing I initially wanted the command to be one line and just kept it as it is.

firmware/Makefile
```diff
@@ -10,6 +11,20 @@ build: FORCE
 clean: FORCE
 	rm $(BUILD_DIR) -rf
 
+deepclean: clean
+ifeq ($(findstring stm32,$(PLATFORM)),stm32)
+	@echo "Deleting files and directories in $(CUBEMX_DIR) that are ignored by Git"
+	@cd $(CUBEMX_DIR) && \
+		for file in $$(ls -R); do \
```
BlakeFreer, line 18: see note in teams, the `find` command should work in git bash

firmware/Makefile
```diff
@@ -10,6 +11,20 @@ build: FORCE
 clean: FORCE
 	rm $(BUILD_DIR) -rf
 
+deepclean: clean
+ifeq ($(findstring stm32,$(PLATFORM)),stm32)
+	@echo "Deleting files and directories in $(CUBEMX_DIR) that are ignored by Git"
+	@cd $(CUBEMX_DIR) && \
```
BlakeFreer, line 17: we do not need the command to be a one-liner, please split into separate commands (remove the `&& \`)

BlakeFreer (changes requested):

BlakeFreer (approved):

PR #66, opened 2024-03-15
User/freerb/fc

BlakeFreer:
Creates the FrontController project along with some basic implementation of the app-level objects.

Adds `vehicle_control_system` as a submodule in the FrontController project and adds it to the project's CMake. Builds successfully.

Creates a `ControlSystem` class that handles initializing and updating the simulink model. This is provides in `inc/simulink.h`. Some of the simulink struct types are aliased to allow the main program to prepare "buffered" instances of the input before setting the control system's actual parameters.

This project is nowhere near complete, but it needs freertos and CAN soon so I want to add it to the main branch.


firmware/projects/DemoProject/platforms/stm32f767/cubemx/board_config.ioc
BlakeFreer, line 1: updates the version to 6.8.1 as listed in the firmware/readme

firmware/projects/FrontController/inc/simulink.h
```diff
@@ -0,0 +1,29 @@
...
+ * For now, since there is only one model, this non-shared implemenation is
+ * fine.
+ */
+class ControlSystem {
+public:
+    static void Initialize() {
+        controller_autogen_initialize();
+    }
+
+    static SimulinkOutput Update(SimulinkInput input) {
```
samparent97, line 24: nice

firmware/projects/FrontController/inc/app.h
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+private:
+    uint16_t position_ = 0;
+    shared::periph::ADCInput& adc_;
+};
+
+class Speaker {
+public:
+    bool state;
```
samparent97, line 57: Should we modify this through an accessor like `SetState` to more closely match the style of `SteeringWheel` for example

firmware/projects/FrontController/inc/simulink.h
```diff
@@ -0,0 +1,29 @@
...
+typedef ExtY SimulinkOutput;
+
+/**
+ * @brief Provides an interface to the simulink model.
+ * @todo Can this become a shared component? Maybe templated off of the i/o
+ * struct types. Such an implementation would also need the initialize() and
+ * step() functions bound in the project, since they could not be referenced
+ * without a project-specific include.
+ * For now, since there is only one model, this non-shared implemenation is
+ * fine.
```
samparent97, line 16: agreed

firmware/projects/FrontController/inc/simulink.h
```diff
@@ -0,0 +1,29 @@
+/// @author Blake Freer
+/// @date 2024-03-14
+
+#include "../vehicle_control_system/src/vehicle_control_system-autogen/controller_autogen_ert_rtw/controller_autogen.h"
```
samparent97, line 4: Is the relative parent directory in the file path needed?

firmware/projects/FrontController/main.cc
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * @brief Copy values to the simulink input struct.
+ */
+void SetControlSystemInputs_Task() {
+    simulink_buffer_output = ControlSystem::Update(simulink_buffer_input);
+}
+
+void ControllerAutogenStep_Task() {}
```
samparent97, line 66: Why is this emtpy?

firmware/projects/FrontController/platforms/stm32f767/cubemx/README
```diff
@@ -0,0 +1,5 @@
+All files generated by CubeMX will be placed here.
+
+The .ioc file is the only one tracked under version control. It contains all the necessary information to generate the other files.
+
+Before building the project, you are expected to open the .ioc and click 'Generate Code' locally.
```
samparent97, line 5: Update comment

firmware/projects/FrontController/main.cc
BlakeFreer: With the `output = Update(input)` setup, it doesn't make sense to have separate SetInput(), Step(), and GetOutput() tasks since they all happen in one. This function can be removed

firmware/projects/FrontController/platforms/stm32f767/cubemx/README
BlakeFreer: I will make this an issue since it needs to be applied to all projects

firmware/projects/FrontController/inc/simulink.h
BlakeFreer: I should check, you're right that the CMake include directories should handle this

firmware/projects/FrontController/inc/app.h
BlakeFreer: depends on how we want our Update to work. If SetState updates the GPIO then yes I agree, but if we require a separate Update function (for example, to update all GPIO in the same task), then I don't see the benefit of hiding a boolean variable. This is a slightly different setup than SteeringWheel since Speaker / BrakeLight are outputs, not inputs.

firmware/projects/FrontController/inc/simulink.h
BlakeFreer: see latest commit, this path has been shortened

firmware/projects/FrontController/main.cc
samparent97: Yea lets remove it in that case

samparent97:
```suggestion
```

samparent97 (approved):

firmware/projects/FrontController/main.cc
BlakeFreer, line 1: I removed that function using Github, so clang-format was not applied

.vscode/c_cpp_properties.json
BlakeFreer, line 1: removes some unnecessary include paths and adds one for simulink. I am working on a better intellisense system since not all projects will have a simulink folder, but this does not negatively affect other projects

samparent97 (approved):

PR #75, opened 2024-03-22
User/freerb/gtest

### LINKED ISSUE #53: Unit Tests!!! - Mappers
Write tests for the various mapper functions. Try to break them.

See racecar/firmware/shared/util/mappers/

BlakeFreer:
Closes #53 

Adds a tests folder, includes googletest as a submodule, and writes tests for all mappers

The build system needs some work, but this is fine for now

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "firmware/projects/FrontController/vehicle_control_system"]
 	path = firmware/projects/FrontController/vehicle_control_system
 	url = https://github.com/macformula/vehicle_control_system.git
+[submodule "firmware/tests/googletest"]
+	path = firmware/tests/googletest
+	url = https://github.com/google/googletest.git
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -49,9 +49,12 @@
             "name" : "Test",
             "includePath": [
                 "${workspaceFolder}/firmware",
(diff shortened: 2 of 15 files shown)


firmware/tests/googletest
samparent97, line 1: We should add google test to the third party directory.

BlakeFreer: I'd prefer to leave it under tests for now since the build system is separated. I agree that it should move to 3rd party in the future but rhat will be a bit messy with the current CMake

samparent97: I think we had a agreed a couple weeks back to keep third-party folder in the firmware folder

BlakeFreer: yes we did, although that conversation did not included testing and we didn't know that a different build system would be needed for the tests. I can continue trying to get the CMake to build out-of-tree code but the cleanest solution on that front is to leave googletest as subdirectory of tests/. 

Once the build systems have been unified, it is a trivial change to move googletest into fw/third-party, so we are not incurring any technical debt.

BlakeFreer: I worked on this for another hour and a half with no success. Moving googletest out of the subdirectory tree screws up the build process, so it will be left here for now

samparent97 (approved):
Good for now, lets add an issue describing this though.

BlakeFreer:
See #84

PR #77, opened 2024-03-22
deepclean bugfix

BlakeFreer:
The `cd CUBEMX_FOLDER` has no effect in the Makefile deepclean target, so deepclean was attempting to remove all non-git tracked files, not just those in the specified cubemx directory. This PR fixes it by explicitly telling `find` which directory to look in

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -14,8 +14,7 @@ clean: FORCE
 deepclean: clean
 ifeq ($(findstring stm32,$(PLATFORM)),stm32)
 	@echo "Deleting files and directories in $(CUBEMX_DIR) that are ignored by Git"
-	@cd $(CUBEMX_DIR)
-	@find | xargs git check-ignore | xargs rm -rf
+	find $(CUBEMX_DIR) | xargs git check-ignore | xargs rm -rf
 else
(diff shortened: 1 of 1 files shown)


firmware/Makefile
```diff
@@ -14,8 +14,7 @@ clean: FORCE
 deepclean: clean
 ifeq ($(findstring stm32,$(PLATFORM)),stm32)
 	@echo "Deleting files and directories in $(CUBEMX_DIR) that are ignored by Git"
-	@cd $(CUBEMX_DIR)
-	@find | xargs git check-ignore | xargs rm -rf
+	find $(CUBEMX_DIR) | xargs git check-ignore | xargs rm -rf
```
Micnasr, line 17: Would suppressing the find command from printing to the terminal be better to keep it clutter free or no?

BlakeFreer: The command doesn't have any console output, so I think it's nice to show what is being executed

tullom: Agree with Blake, the added verbosity is nice to have since the command doesn't output anything to the console

tullom (approved):
LGTM

PR #59, opened 2024-03-10
Added support for FreeRTOS and added FreeRTOS example

tullom:
Added support for FreeRTOS in our firmware abstracted through an OS layer. An example project is available to be built, `DemoProject` for STM32F767.

- Added `shared::util::os` class that all OS layer objects inherit from
- Added `osStatus` enum class to define return types
- Added full support for semaphores, mutexes, FIFOs, software timers, and tick (delay functionality), and partial support for event flags
- Added full support for the above `osDataType` classes to be used with STM32F767 using `mcal::os`, except event flags
- Ported DemoProject to use FreeRTOS as an example of how to use the newly supported data structures
- Modified DemoProject `board_config.ioc` to generate FreeRTOS files.

**DemoProject functionality description**
- Three FreeRTOS tasks are created
  - `StartTestTask` - this task polls a button at 1kHz, when pressed a semaphore `sem_test` is signaled
  - `StartLedTask` - this task tries to acquire a semaphore `sem_test`. When the button is pressed, the semaphore is signaled and `StartLedTask` toggles the red LED
  - `StartMessageSendTask` - this task starts a one-shot software timer `timer_test`. The task then starts polling a FIFO `message_queue_test` for new messages to copy to its local buffer. Once the timer callback `messageTimerCallback` fires after 3 seconds, the timer writes the contents of a string buffer to the FIFO. Once `StartMessageSendTask` sees that the entire string has been copied, it checks to make sure the received string is the same as the original string. If true, the blue LED is set.

PR #82, opened 2024-03-25
RTOS Update

BlakeFreer:
Resolves comments from PR #59


samparent97 (approved):

samparent97:
While I would like to get this merged, I want to point out that it is a bit strange that the timer is not a child class of periph.

tullom:
> While I would like to get this merged, I want to point out that it is a bit strange that the timer is not a child class of periph.

@samparent97 its because the timer class is a software timer handled by FreeRTOS and its own tick system, not a hardware timer. Indeed, it might be wise to rename the timer class to SoftTimer or something like that.

firmware/mcal/stm32f767/os/timer.h
```diff
@@ -5,33 +5,33 @@
...
-        int ret = osTimerStop(*timer_id);
-        return ret == 0 ? shared::util::OsStatus::kOsOk : shared::util::OsStatus::kOsError;
+    OsStatus Stop() override {
+        auto ret = osTimerStop(*timer_id_);
+        return ret == osOK ? OsStatus::kOk : OsStatus::kError;
     }
-    uint32_t IsRunning() override {
-        return osTimerIsRunning(*timer_id);
+    bool IsRunning() override {
+        return osTimerIsRunning(*timer_id_);
```
tullom, line 30: Nit: i did not know C++ could implicitly cast this to a bool. if it works, it works, but it might make things clearer if you `static_cast<bool>` this

firmware/projects/DemoProjectRTOS/README.md
```diff
@@ -1,71 +1,3 @@
...
-This will delete the `build/DemoProject/<platform>` folder.
-
-Optionally, you can omit the `PLATFORM` flag to delete the entire `build/DemoProject` folder, or omit both flags and simply call `make clean` to delete all build files for all projects and platforms.
-
-The `build` folder is not tracked by git so you can delete it anytime.
-
-## Troubleshooting
-
-Ensure you have all dependencies installed. See `racecar/firmware/README.md` for a complete list of dependencies.
+Demonstrates a simple button program using FREERTOS tasks.
```
tullom, line 3: Maybe pull the description i used for my pull request here: https://github.com/macformula/racecar/pull/59

firmware/shared/os/timer.h
```diff
@@ -4,15 +4,16 @@
 #pragma once
 
 #include <cstdint>
-#include "shared/util/os.h"
+
+#include "shared/os/os.h"
 
 namespace shared::os {
 
 class Timer {
```
tullom, line 12: based on discussion, change to `SoftTimer`

tullom: actually nevermind, this can be the base class for both the software timer and the hardware timer (when we implement it)

firmware/mcal/stm32f767/os/timer.h
```diff
@@ -5,33 +5,33 @@
...
+#include "shared/os/timer.h"
 
 namespace mcal::os {
 
-
 // TODO: Add comments and handle errors more robustly
-class Timer final : public shared::os::Timer  {
-private:
-    osTimerId_t* timer_id;
+class Timer final : public shared::os::Timer {
```
tullom, line 15: based on discussion, change to `SoftTimer`

BlakeFreer: SoftTimer as in SoftwareTimer? If so can I use the full word to make it more clear?

firmware/mcal/stm32f767/os/timer.h
BlakeFreer: yes, zero will cast to `false` and anything else is `true`, but the `static_cast` is a great idea

firmware/mcal/stm32f767/os/timer.h
BlakeFreer: I realized that the name is already quite descriptive since the namespace makes it `os::Timer` which perfectly describes its origin and purpose

firmware/projects/DemoProjectRTOS/README.md
BlakeFreer: done

firmware/mcal/stm32f767/os/timer.h
BlakeFreer: done

firmware/mcal/stm32f767/os/timer.h
tullom: makes perfect sense. no change needed

tullom (approved):

PR #83, opened 2024-03-25
User/freerb/can changes

BlakeFreer:


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "third-party/etl"]
+[submodule "firmware/third-party/etl"]
 	path = firmware/third-party/etl
-	url = git@github.com:ETLCPP/etl.git
+	url = https://github.com/ETLCPP/etl.git
```
BlakeFreer, line 3: https cloning is more standard than ssh

firmware/mcal/raspi/periph/can.h
```diff
@@ -1,74 +1,32 @@
...
-#ifndef MCAL_RASPI_PERIPH_CAN_H_
-#define MCAL_RASPI_PERIPH_CAN_H_
+#pragma once
 
-#include <cstdint>
-#include <iostream>
-#include <cstdint>
-#include <cstring>
-#include <queue>
+#include <linux/can.h>
```
BlakeFreer, line 6: clang-format sorts includes

firmware/mcal/raspi/periph/can.h
```diff
@@ -123,8 +83,50 @@ class CanBase : public shared::periph::CanBase {
 
         return;
     }
-};
 
-} // namespace mcal::periph
+private:
```
BlakeFreer, line 87: moves private section after public

firmware/projects/DemoCan/main.cc
```diff
@@ -15,7 +15,7 @@ extern mcal::periph::CanBase veh_can_base;
 extern void Initialize();
 }  // namespace bindings
 
-generated::can::DemoCanVehMsgRegistry veh_can_regitry{};
+generated::can::DemoCanVehMsgRegistry veh_can_registry{};
```
BlakeFreer, line 18: typo

BlakeFreer (commented):
.

.gitmodules
samparent97: Never realized git modules differentiated between the two. With the introduction of PAT with GitHub https repo URL's, I believe ssh has become the standard. Nonetheless I don't think this matters much since we are not trying to push changes via the submodule, we just want to clone

samparent97 (approved):

.gitmodules
BlakeFreer: I had authentication issues when trying to initialize the submodule. I probably could have solve it with more work but the issue is avoided with https

PR #92, opened 2024-04-03
Switches to clangd as a language server

### LINKED ISSUE #58: Environment configuration with Makefile
We do not need to use c_cpp_properties.json. The clangd extension (by LLVM) is an alternative

## Original Issue

To use Intellisense in VSCode, you must tell the IDE of which project and platform you are developing for.

Currently, the Platform is selected with the `>C/C++: Select a Configuration...` function, while the Project is selected by modifying the "env.PROJECT" variable in [.vscode/c_cpp_properties.json](https://github.com/macformula/racecar/blob/main/.vscode/c_cpp_properties.json#L3)

1. Add a recipe to the master Makefile called `config-env` which sets some environment variables for PLATFORM and PROJECT.
2. Reference the variables within the `c_cpp_properties.json` file to replace the old built-in variable

This should be "easy" for the project since it is an explicit variable in the json file. I do not know how to programmatically change the configuration platform (ex windows vs stm). It is ok if you can only get it to work for the project as this is the more tedious one to change manually.

BlakeFreer:
# **clangd**

Closes #58

Our previous language server was the Microsoft **C/C++** extension for VSCode, which had several limitations:

1. Inability to easily update configurations
    - You had to frequently change the c_cpp_properties settings
2. Disconnect between server and build system
    - We had to manually specify all of the include paths for each project, but these were not uniform across projects.
    - The IDE's server may not match the build settings, causing hard-to-detect errors when include paths were missing
    - Updating the build system required updating the config
3. Only works for VSCode, and the config had to be under version control

**clangd** solves all of these problems

1. The language server settings (`compile_commands.json`) are automatically generated every time a project is built with the Makefile
2. Since CMake generates the compile_commands file, the language server knows exactly how the project is being built.
    - This means the CMakeLists files are the single source of truth
3. Multiple IDEs support clangd

## Installation

See instructions in `firmware/README.md`


firmware/projects/DemoMapper/platforms/windows/CMakeLists.txt
```diff
@@ -1,5 +1,3 @@
-add_library(bindings)
```
samparent97, line 1: Could you explain this change, I see it has been applied to all files.

firmware/CMakeLists.txt
```diff
@@ -54,21 +54,23 @@ set(CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTOR
 set(CMAKE_CXX_EXTENSIONS ON)
 
 project("${PROJECT}-${PLATFORM}")
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+set(CMAKE_C_USE_RESPONSE_FILE_FOR_INCLUDES OFF) # this doesn't seem to do anything yet
 
 add_executable(main)
+add_library(bindings)
```
samparent97, line 61: I see the answer to my previous comment now.

firmware/projects/DemoMapper/platforms/windows/CMakeLists.txt
samparent97: The add library for bindings got moved to the higher level cmake.

samparent97 (approved):
This is a phenomenal change. The switch to the clangd language server not only allows us to remove environment specific files (.vscode folder) but also let's us remove the large drivers folder we used for symbol resolution. Great work.

samparent97:
> This is a phenomenal change. The switch to the clangd language server not only allows us to remove environment specific files (.vscode folder) but also let's us remove the large drivers folder we used for symbol resolution. Great work.

It is also blazing fast.

PR #100, opened 2024-04-14
build: :building_construction: Revamp the build system. 💯 

### LINKED ISSUE #42: Update toolchain to 13.2
1. Change firmware/README
2. Add a "retarget" file for stm32 to provide functions like _read and _close. See [this header file from TMS GEN 1](https://github.com/macformula/thermal_monitoring/blob/main/TMS_GEN1/Core/Inc/retarget.h) and its [associated source file](https://github.com/macformula/thermal_monitoring/blob/main/TMS_GEN1/Core/Src/retarget.c). Include the file into CMake (likely under cubemx -> drivers target)

### LINKED ISSUE #80: Update bindings CMake
determine if this can be made private since `bindings.h` no longer exists

```cmake
# must be public so that the 'main' executable can see its headers when including bindings.h
target_link_libraries(bindings PUBLIC driver)
```

if change can be made, apply to all projects

### LINKED ISSUE #81: Update the build system
The current architecture includes header files that contain mcal / shared code. This is not standard since code is implemented in header files. A better method would be the common header .h and implementation .cc structure

This means that the shared/mcal libraries will each have their own CMake that defines a library (ex `mcal-periph` or `shared-util`) that can be linked to the `main` target during the build process. This will modularize compilation and speed up building since common code will not need to be re-compiled

This will also enable more interesting code structures since implementations can be switched out at compile time, rather than relying on static code. ~~For example, we may be able to remove the `shared` and `mcal` namespaces since mcal peripheral code will be placed in a `.cc` file that provides the implementation for a shared `.h` file.~~
> This example is not possible as it would require all platforms' implementation of a peripheral to have the same instance variables and non-interface methods, which would make it impossible to implement for different platforms, but we should still pursue this issue for the other reasons.

BlakeFreer:
Convert mcal and shared into CMakeLibraries, allowing source files in the future. Add mcal_conf to each project platform to explicity state which mcal is being used. Extract variables from cubemx Makefile to allow generic build scripts.

Closes #42 
Closes #81
Closes #80


firmware/mcal/windows/periph/CMakeLists.txt
samparent97, line 1: add a coment

samparent97 (approved):

Issue #69, opened 2024-03-19
GitHub action for building code


BlakeFreer:
When a Pull Request is created, all projects and platforms should be built to ensure there are no errors. We can accept warnings for now, especially until #42 is completed to remove the "not implemented" warnings.

This means running `make PROJECT=<proj> PLATFORM=<plat>` for all projects and each projects' supported platforms and rejecting the PR if there are any errors. Since the projects and platforms do not change frequently, I am fine with manually defining these commands.

If the action takes a long time to run, see if there is a way to only run the tests for the projects that are changed in new PR.

The action should use a Linux machine since it is the cheapest option available. See `firmware/README` for the machine requirements and make sure to use the exact SW versions.

An extension would be automatically parsing the `firmware/projects` folder to find all projects, and the `firmware/projects/<proj>/platforms` folder for each project, to define these generate these commands automatically.

This SO post may help https://stackoverflow.com/questions/58654530/how-to-auto-reject-a-pull-request-if-tests-are-failing-github-actions

BlakeFreer:
Completed by #419

PR #110, opened 2024-04-16
STM32 Can implementation; DemoCan Example

samparent97:


firmware/cmake/generate_cubemx.mk
BlakeFreer, line 1: This will not work. The `Makefile` target will not be executed. CMake calls `make` which will build the first available target, which is GenerateCubeMx so the cubemx Makefile will never be updated with the custom targets.

I don't understand why this file has to change at all. The generation and Makefile modification are inherently coupled and should stay that way.

firmware/cmake/generate_cubemx.mk
```diff
@@ -34,4 +37,4 @@ Makefile: $(IOC_FILE)
 # This is used by build_cubemx.cmake
 	echo "" >> Makefile
 	echo "%.value:" >> Makefile
-	echo -e "\t@echo \$$(\$$*)" >> Makefile
\ No newline at end of file
+	echo "\t@echo \$$(\$$*)" >> Makefile
```
BlakeFreer, line 40: I don't know why I had this

BlakeFreer (changes requested):

firmware/cmake/generate_cubemx.mk
BlakeFreer: https://www.gnu.org/software/coreutils/manual/html_node/echo-invocation.html#echo-invocation

Without `-e`, echo will print `\t` instead of a tab character

firmware/cmake/generate_cubemx.mk
```diff
@@ -14,17 +14,20 @@ CUBEMX_GEN_SCRIPT = cubemx_script.txt
 # Makefile, i.e. whenever IOC_FILE has been updated but new code has not been
 # generated.
 
-Makefile: $(IOC_FILE)
-# Create an file containing commands to generate the cubemx code.
-	echo "config load \"$(IOC_FILE)\"" > $(CUBEMX_GEN_SCRIPT)
-	echo "project generate ./" >> $(CUBEMX_GEN_SCRIPT)
-	echo "exit" >> $(CUBEMX_GEN_SCRIPT)
+.PHONY: Makefile
```
BlakeFreer, line 17: This should not be PHONY: Makefile is a real file target

BlakeFreer (changes requested):

firmware/cmake/generate_cubemx.mk
samparent97: This is the output when adding -e on my end:
```Make
%.value:
-e 	@echo $($*)
```

Without -e:
```Make
%.value:
	@echo $($*)
```

It is likely this is a unix vs windows thing.

samparent97: This line also gives me correct output, could you give it a try on your end?
```Bash
echo '\t@echo $$($$*)' >> Makefile
```

firmware/cmake/generate_cubemx.mk
samparent97: Will change this.

firmware/cmake/generate_cubemx.mk
samparent97: Given we are not producing the file and simply modifying it, would we consider that to be a real file target?

firmware/cmake/generate_cubemx.mk
samparent97: What we really need here is to get builds working in regressions. @Micnasr do you have time to help us out with this?

BlakeFreer: > This line also gives me correct output, could you give it a try on your end?
> 
> ```shell
> echo '\t@echo $$($$*)' >> Makefile
> ```

This does not work in git-bash. I have a PR to this branch to fix all of this

Micnasr: @samparent97 I am looking at the PR and the changes but I am struggling to understand what we are trying to achieve here and what the problem is

samparent97: @Micnasr slightly unrelated to this PR. But if you could get builds running as a github action to block merge that would be great. Check out issue #69

firmware/cmake/generate_cubemx.mk
samparent97: Getting solved in #111

firmware/mcal/raspi/PostBuild.cmake
BlakeFreer, line 1: why is this all commented out?

firmware/mcal/stm32f767/periph/can.h
```diff
@@ -0,0 +1,142 @@
...
+
+        raw_rx_msg.header = unpack_stm_rx_header(stm_rx_header);
+
+        can_queue_.Enqueue(raw_rx_msg);
+    }
+
+private:
+    static constexpr uint8_t kMaxMsgBytes = 8;
+    static constexpr size_t kMsgQueueLen = 20;
+    static constexpr uint32_t kCanRxFifo0 = CAN_RX_FIFO0;
```
BlakeFreer, line 82: is this a HAL constant? just checking

firmware/projects/DemoCanBar/main.cc
```diff
@@ -0,0 +1,47 @@
...
+shared::can::CanBus veh_can_bus{
+    bindings::veh_can_base,
+    veh_can_registry,
+};
+
+int main(void) {
+    bindings::Initialize();
+    uint32_t tick_duration = 100;
+
+    generated::can::TempSensors temp_sens_msg{};
```
BlakeFreer, line 27: why do these need a namespace? or can we do `using namespace generated::can`?

firmware/projects/DemoCanBar/main.cc
```diff
@@ -0,0 +1,47 @@
...
+    generated::can::TempSensorsReply temp_sens_msg_reply{};
+
+    while (true) {
+        bindings::TickBlocking(tick_duration);
```
BlakeFreer, line 31: I think its more common for the delay to be at the end of the loop

firmware/projects/DemoCanBar/platforms/stm32f767/.DS_Store
BlakeFreer, line 1: remove this file

firmware/projects/DemoCanFoo/README.md
```diff
@@ -0,0 +1,14 @@
+# DemoCan Project
+
+This DemoCan project is best suited for the `raspi` and the `stm32f767` platform. It sends and receives some dummy messages created for testing purposes.
+
+### Nodes:
+`Foo` (this ecu)\
```
BlakeFreer, line 6: use double newline instead of backslash

firmware/projects/DemoCanFoo/README.md
```diff
@@ -0,0 +1,14 @@
...
+`Bar` (some other ecu)
+
+### Messages:
+
+`Foo` sends `Bar` the `TempSensors` message\
```
BlakeFreer, line 11: as above

firmware/projects/DemoCanFoo/main.cc
```diff
@@ -0,0 +1,45 @@
...
+};
+
+int main(void) {
+    bindings::Initialize();
+    uint32_t tick_duration = 100;
+
+    generated::can::TempSensors temp_sens_msg{};
+
+    while (true) {
+        bindings::TickBlocking(tick_duration);
```
BlakeFreer, line 30: move to end of loop

firmware/projects/DemoCanFoo/platforms/raspi/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+target_sources(bindings
+	PRIVATE
+	bindings.cc
+)
+
+target_include_directories(bindings PUBLIC ${CMAKE_SOURCE_DIR})
```
BlakeFreer, line 6: this should be replaced with 

```cmake
target_link_libraries(bindings PUBLIC mcal-raspi)
```

firmware/projects/DemoCanBar/platforms/raspi/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+target_sources(bindings
+	PRIVATE
+	bindings.cc
+)
+
+target_include_directories(bindings PUBLIC ${CMAKE_SOURCE_DIR})
```
BlakeFreer, line 6: replace with 

```cmake
target_link_libraries(bindings PUBLIC mcal-raspi)
```

firmware/shared/util/data_structures/circular_queue.h
BlakeFreer, line 1: nice

BlakeFreer (changes requested):

firmware/mcal/stm32f767/periph/can.h
samparent97: Yes

firmware/projects/DemoCanBar/main.cc
samparent97: There will be many can messages generated and they will be based on dbc. I don't want those to pollute namespace.

firmware/projects/DemoCanBar/platforms/stm32f767/.DS_Store
samparent97: 😑

BlakeFreer (approved):

PR #111, opened 2024-04-20
build: :construction_worker: Improve the CubeMX autogeneration code.

BlakeFreer:
Provide a build option to bypass CubeMX generation (set AUTOGEN_CUBEMX=OFF in the top-level make command). Change method for appending custom targets to the cubemx Makefile to use a file copy instead of string printing.


firmware/cmake/generate_cubemx.mk
```diff
@@ -7,34 +7,31 @@ else
...
-# 	java -jar "$(CUBEMX_PATH)" -q "$(CUBEMX_GEN_SCRIPT)"
-
-# 	rm $(CUBEMX_GEN_SCRIPT)
-
 Makefile: $(IOC_FILE)
-# Add a recipe for building the sources to objects without linking them.
-# This is used by build_cubemx.cmake
-	echo "" >> Makefile
-	echo 'objects: $$(OBJECTS)' >> Makefile
+	@echo "Autogenerating from CubeMX. If you don't want do this, you must manually 'Generate Code' before building."
```
samparent97, line 28:
```suggestion
	@echo "Autogenerating from CubeMX. If you don't want to do this, you must manually 'Generate Code' before building."
```

PR #122, opened 2024-04-29
Fix CAN Rx interrupts for stm32f767 platform

### LINKED ISSUE #112: Fix CAN Rx interrupts for stm32f767
CAN Rx interrupts are currently not firing for project `DemoCanBar` as we would expect them to ,

samparent97:
closes #112


firmware/mcal/stm32f767/periph/can.h
```diff
@@ -83,6 +84,19 @@ class CanBase : public shared::periph::CanBase {
     static constexpr uint32_t kCanRxActiveInterruptFifo0 =
         CAN_IT_RX_FIFO0_MSG_PENDING;
 
+    static constexpr uint32_t kDefaultFilterIdHigh = 0x0000;
+    static constexpr uint32_t kDefaultFilterIdLow = 0x0000;
+    static constexpr uint32_t kDefaultFilterMaskIdHigh = 0x0000;
+    static constexpr uint32_t kDefaultFilterMaskIdLow = 0x0000;
+
+    static constexpr uint32_t kFilterScale = CAN_FILTERSCALE_32BIT;
+    static constexpr uint32_t kFilterEnable = CAN_FILTER_ENABLE;
```
BlakeFreer, line 93: What's the purpose of renaming constants? Is it to add ownership of a value to this class?

BlakeFreer (commented):
Seems good

firmware/mcal/stm32f767/periph/can.h
tullom: I believe Sam is doing this to signify a separation between the underlying cubemx driver code and our app layer code. Plus it follows the google naming standard for variables so I think this is a good change.

firmware/mcal/stm32f767/periph/can.h
```diff
@@ -21,8 +21,9 @@ class CanBase : public shared::periph::CanBase {
     CanBase(CAN_HandleTypeDef* hcan) : hcan_(hcan){};
 
     void Setup() {
-        HAL_CAN_ActivateNotification(hcan_, kCanRxActiveInterruptFifo0);
+        ConfigFilters();
         HAL_CAN_Start(hcan_);
+        HAL_CAN_ActivateNotification(hcan_, kCanRxActiveInterruptFifo0);
```
tullom, line 26: are there any implications to adding this line after `HAL_CAN_Start()`? Even if not, i think it should be moved above to be consistent with how it was before.

tullom (changes requested):
@samparent97 back to you

firmware/mcal/stm32f767/periph/can.h
samparent97: Did this based on some old code that had working rx interrupts. But I do believe it works either way. I'll test it out and will make the change if functional.

samparent97: Looks like it works 👍

PR #123, opened 2024-04-30
Add BMS Broadcast to tms project

### LINKED ISSUE #120: Fix warnings in CAN Templates
The current CAN templates cause float-conversion and suggest-override compiler warnings

```
warning: conversion from 'double' to 'float' changes value from '1.0000000000000001e-1' to '1.00000001e-1f' [-Wfloat-conversion]
   39 |     static constexpr float kAMK_TempMotorScale = 0.1;

warning: 'virtual bool generated::can::VehMsgRegistry::SetMessage(const shared::can::RawCanMsg&)' can be marked override [-Wsuggest-override]
   18 |     bool SetMessage(const shared::can::RawCanMsg& raw_msg) {
```

Edit the templates to avoid these warnings.

samparent97:
Adds CAN message `BmsBroadcast` to the TMS project

closes #120


firmware/dbcs/VEH_CAN.dbc
```diff
@@ -79,15 +79,15 @@ BO_ 1574 Contactor_Feedback: 3 BMS
...
+BO_ 2553934720 BmsBroadcast: 8 TMS
  SG_ ThermModuleNum : 0|8@1+ (1,0) [0|0] ""  BMS
  SG_ LowThermValue : 8|8@1- (1,0) [0|0] " C"  BMS
  SG_ HighThermValue : 16|8@1- (1,0) [0|0] " C"  BMS
  SG_ AvgThermValue : 24|8@1- (1,0) [0|0] " C"  BMS
  SG_ NumThermEn : 32|8@1+ (1,0) [0|0] ""  BMS
  SG_ HighThermID : 40|8@1+ (1,0) [0|0] ""  BMS
  SG_ LowThermID : 48|8@1+ (1,0) [0|0] ""  BMS
- SG_ Checksum : 56|8@1- (1,0) [0|0] ""  BMS
+ SG_ Checksum : 56|8@1+ (1,0) [0|0] ""  BMS
```
samparent97, line 90: checksum is meant to be unsigned.

firmware/projects/TMS/inc/app.h
```diff
@@ -151,4 +156,43 @@ class DebugIndicator {
...
+private:
+    static constexpr uint8_t kThermistorModuleNumber = 0;
+
+    shared::can::CanBus& can_bus_;
+    generated::can::BmsBroadcast bms_broadcast_;
+    uint8_t num_thermistors_;
+
+    uint8_t CalculateBmsBroadcastChecksum(
+        const generated::can::BmsBroadcast& bms_broadcast) {
+        constexpr int8_t kChecksumConstant = 0x41;
```
samparent97, line 190: This is a constant defined by Orion 2. They don't actually tell you about it, we had to reverse engineer the TMS expansion pack to uncover it.

firmware/projects/TMS/platforms/stm32f767/CMakeLists.txt
```diff
@@ -3,5 +3,5 @@
 
 add_subdirectory(cubemx)
 
-target_sources(bindings PUBLIC bindings.cc)
+target_sources(bindings PUBLIC bindings.cc os.cc)
```
samparent97, line 6: We weren't compiling the os.cc file in mainline??

scripts/cangen/templates/msg_registry.h.jinja2
```diff
@@ -6,17 +6,17 @@
...
 #include "shared/comms/can/raw_can_msg.h"
-#include "etl/unordered_map.h"
 
 namespace generated::can {
 
 {% set reg_class_name = bus_name + "MsgRegistry" %}
 class {{ reg_class_name }} : public shared::can::MsgRegistry {
 public:
-    bool SetMessage(const shared::can::RawCanMsg& raw_msg) {
+    bool SetMessage(const shared::can::RawCanMsg& raw_msg) override {
```
samparent97, line 19: Removes clangd warnings on generated files.

firmware/projects/TMS/inc/app.h
BlakeFreer: can you add this explanation as a comment on the constant?

firmware/projects/TMS/inc/app.h
```diff
@@ -151,4 +153,45 @@ class DebugIndicator {
...
+
+        can_bus_.Send(bms_broadcast_);
+    }
+
+private:
+    static constexpr uint8_t kThermistorModuleNumber = 0;
+
+    shared::can::CanBus& can_bus_;
+    generated::can::BmsBroadcast bms_broadcast_;
+    uint8_t num_thermistors_;
```
BlakeFreer, line 185: Why does `num_thermistors` need to be an instance variable? It seems like it would fit in better as an argument to `SendBmsBroadcast()`. I know this means that we repeatedly pass the value, but it reduces coupling at construction time, and at call time we will already be handling the temperature arrays so the number will be readily available.

firmware/projects/TMS/inc/app.h
```diff
@@ -151,4 +153,45 @@ class DebugIndicator {
...
+
+    uint8_t CalculateBmsBroadcastChecksum(
+        const generated::can::BmsBroadcast& bms_broadcast) {
+        constexpr int8_t kChecksumConstant = 0x41;
```
BlakeFreer, line 189: should this be a uint8? not sure if it makes a difference in the end

firmware/projects/TMS/main.cc
```diff
@@ -1,12 +1,18 @@
 /// @author Blake Freer
 /// @date 2023-11-18
 
+#include <sys/_stdint.h>
```
BlakeFreer, line 4: remove this, clangd likes to add it but cstdint provides the necessary defs

firmware/projects/TMS/main.cc
```diff
@@ -113,26 +132,33 @@ TempSensor temp_sensors[] = {
...
+    static uint8_t high_thermistor_idx;
 
+    veh_can_bus.Update();
     ts_manager.Update();
     ts_manager.GetTemperatures(temperature_buffer);
 
-    float temp_min =
-        shared::util::GetMinimum<float, kSensorCount>(temperature_buffer, NULL);
+    float temp_min = shared::util::GetMinimum<float, kSensorCount>(
+        temperature_buffer, &low_thermistor_idx);
```
BlakeFreer, line 151: nice

firmware/projects/TMS/main.cc
```diff
@@ -113,26 +132,33 @@ TempSensor temp_sensors[] = {
...
+    float temp_max = shared::util::GetMaximum<float, kSensorCount>(
+        temperature_buffer, &high_thermistor_idx);
 
     float temp_avg =
         shared::util::GetAverage<float, kSensorCount>(temperature_buffer);
 
-    /// TODO: Pack & send CAN message
+    bms_broadcaster.SendBmsBroadcast(
+        high_thermistor_idx, static_cast<int8_t>(temp_max), low_thermistor_idx,
+        static_cast<int8_t>(temp_min), static_cast<int8_t>(temp_avg));
```
BlakeFreer, line 161: SInce this is an argument-heavy method, add a trailing comma after the last argument, then clangd will nicely format the  arguments on individual rows. This will help, especially since clangd will then label the arguments.

firmware/projects/TMS/platforms/stm32f767/CMakeLists.txt
BlakeFreer: i have no idea how it was working. I could've sworn we created an `os` CMake library (analogous to the `bindings` library), but I don't see it in TMS or DemoProjectRTOS. I'll have to find where that went

firmware/projects/TMS/platforms/windows/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
-target_sources(bindings PRIVATE bindings.cc)
+target_sources(bindings PRIVATE bindings.cc os.cc)
```
BlakeFreer, line 1: we don't have os for windows. There shouldn't be an os.cc file under platforms/windows

firmware/shared/util/algorithms/arrays.h
```diff
@@ -21,14 +22,14 @@ concept Numeric = requires { std::is_arithmetic_v<T>; };
 /// of the minimum value, the least index is returned.
 /// @return Minimum value of the array.
 /// @note Returns a large value if `array_length = 0`.
-template <Numeric T, int array_length>
-    requires(array_length >= 0)
-T GetMinimum(T* array, int* idx_min) {
+template <Numeric T, int array_length, Numeric I>
```
BlakeFreer, line 25: does the additional template parameter provide any benefit besides checking limits? We really should be using a `size_t` variable for the index return, then the array length limit is unnecessary. It is the responsibility of the programmer to convert the `size_t` value into whatever type they need.

It seems like you added this to help get the `uint8_t high_thermistor_idx` value for TMS. I don't think we should be performing this check for the programmer; they should be aware of the inherent limits of their chosen datatype.

BlakeFreer (changes requested):

firmware/projects/TMS/inc/app.h
samparent97: Made it an instance because we know the number of thermistors at construction time and it will not change over the course of its lifetime. I think this coupling makes sense, thoughts?

firmware/projects/TMS/inc/app.h
samparent97: should be a uint8, yes, no difference in how it will work.

firmware/projects/TMS/main.cc
samparent97: good catch, yea i wonder if theres a way to catch these

firmware/projects/TMS/main.cc
samparent97: I remembered you had told me how to format it but couldnt remember what it was I had to add. Thanks

firmware/shared/util/algorithms/arrays.h
samparent97: why not have the check built in when its easily available? I quite like when we are able to remove the possibility of a mistake with a compile time check.

firmware/projects/TMS/inc/app.h
BlakeFreer: Sure it's fine either way. Template it if you can (like the TMS temp manager) but don't if it gets complicated

PR #125, opened 2024-05-06
Add Front Controller Io Checkout project

samparent97:
Reads all inputs and transmits them over can


firmware/mcal/raspi/periph/can.h
BlakeFreer, line 1: undo these changes, there should not be a trailing space

firmware/mcal/stm32f767/periph/can.h
BlakeFreer, line 1: are all of these settings going to be the same for every CAN? if so, this file is good, otherwise they should not be in mcal/

firmware/projects/DemoCanFoo/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ list(APPEND CAN_DEPENDENCIES "${CMAKE_CURRENT_SOURCE_DIR}/config.yaml")
 
 add_custom_target(
     generated_can
-	COMMAND ${Python3_EXECUTABLE} "${CMAKE_SOURCE_DIR}/../scripts/cangen/main.py" "--project=\"DemoCanFoo\""
+	COMMAND /home/parallels/macformula/racecar/scripts/cangen/.deb_venv/bin/python3.11 "${CMAKE_SOURCE_DIR}/../scripts/cangen/main.py" "--project=\"DemoCanFoo\""
```
BlakeFreer, line 20: revert to the cmake variable, this path is device dependent

firmware/projects/DemoProject/main.cc
```diff
@@ -12,7 +12,7 @@ extern void Initialize();
 Button button{bindings::button_di};
 Indicator indicator{bindings::indicator_do};
 
-bool btn_value;
+bool btn_value = true;
```
BlakeFreer, line 15: this is not necessary since the value is set before it is used

firmware/projects/DemoProject/platforms/sil/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+target_include_directories(bindings
+    PRIVATE
+    "$<TARGET_PROPERTY:mcal-raspi,INTERFACE_INCLUDE_DIRECTORIES>"
```
BlakeFreer, line 3: what does this line do? we have  not needed anything like this before

firmware/projects/DemoProject/platforms/sil/bindings.cc
BlakeFreer, line 1: should the platform folder be called "sil" or "raspi", or should we rename the mcal to "sil", or are they distinct?

firmware/projects/IoCheckoutFc/main.cc
```diff
@@ -0,0 +1,140 @@
...
+            adc_to_voltage.Evaluate(suspension_travel_2_value);
+
+        // Transmit messages
+        io_can_bus.Send(pedal_sensors_msg);
+        io_can_bus.Send(wheels_msg);
+        io_can_bus.Send(vehilcle_info_msg);
+        io_can_bus.Send(suspension_msg);
+
+        // Wait 100ms
+        bindings::TickBlocking(100);
```
BlakeFreer, line 136: I think we should use TickUntil, especially since ADC is relatively slow

firmware/projects/IoCheckoutFc/platforms/sil/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+target_include_directories(bindings
+    PRIVATE
+    "$<TARGET_PROPERTY:mcal-raspi,INTERFACE_INCLUDE_DIRECTORIES>"
```
BlakeFreer, line 3: what is this line?

firmware/shared/util/mappers/linear_map.h
BlakeFreer, line 1: I intentionally did not perform this cast. When higher resolution types (floats, doubles) are mixed with low resolution types (integers), the operations will promote the low-res type to the high-res type, allowing for a more accurate calculation. The rounding only happens after the operation finishes (by casting with `T()`).

Consider when `m=int 10, b= int 0` using U=float, T=int. Without the cast / with promotion, the value 0.1 gets mapped to 1, 0.2 to 2, 1.1 to 11, etc. With the early cast on `x`, 0.1 gets rounded to 0, so the return is 0. Everything up to 0.99 gets rounded down to 0, which is clearly not desired behaviour.

firmware/validation/sil/sil_client.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace val::sil {
+
+class SilClient {
+public:
+    SilClient(std::string server_addr);
+
+    void Connect();
+
+    void RegisterDigitalInput(std::string ecu_name, std::string sig_name);
```
BlakeFreer, line 24: this style will work for now, but this repeated `Register___()` may be solved with OOP on the different peripherals

BlakeFreer (changes requested):

PR #134, opened 2024-06-18
Fix absolute paths issue on Windows

TylerStAmour:
Using realpath and abspath in a makefile that is invoked from another makefile seems to not work on Windows in at least Make 3.8.1 (version included in Git Bash for me).

Using abspath/realpath in this case will append to an existing path, producing a path such as `D:/STWorkspace/racecar/firmware/projects/DemoProject/platforms/stm32f767/cubemx/D:/STWorkspace/racecar/firmware/cmake`.

Haven't tested this fix on Linux since I don't have a setup for it right now, but it should still produce the correct filepath. Not sure how common this issue would be for building on windows but it's only a minor inconvenience anyway.

For bug report: [https://lists.gnu.org/archive/html/bug-make/2008-06/msg00054.html](https://lists.gnu.org/archive/html/bug-make/2008-06/msg00054.html)

### DIFF
--- a/firmware/cmake/generate_cubemx.mk
+++ b/firmware/cmake/generate_cubemx.mk
@@ -12,8 +12,8 @@ AUTOGEN_CUBEMX = ON
 CUBEMX_GEN_SCRIPT = cubemx_script.txt
 
 # Get the directory of this file since it is called from other locations
-BUILD_SYS_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-CUSTOM_TARGETS_FILE = $(BUILD_SYS_DIR)/custom_cubemx_targets.mk
+BUILD_SYS_DIR:=$(dir $(MAKEFILE_LIST))
+CUSTOM_TARGETS_FILE = $(BUILD_SYS_DIR)custom_cubemx_targets.mk
 
 # Copy the CubeMX makefile and custom targets into a new makefile
 CustomMakefile.mk: Makefile $(CUSTOM_TARGETS_FILE)


BlakeFreer:
I made a minimum working example to summarize the setup.
```
C:\
|-- racecar\
|   |-- generate_cubemx.mk
|   |-- inner_dir
|   |   |-- Makefile

[generate_cubemx.mk]
DIR = $(dir $(MAKEFILE_LIST))

all:
	@echo $(DIR)

[inner_dir/Makefile] (replicating firmware/cmake/build_cubemx.cmake)
all:
	@make --file ../generate_cubemx.mk --no-print-directory
```

First, do you think this is an accurate mock-up of the build setup?

With `$(dir $(MAKEFILE_LIST))`, running the inner Makefile prints `../`, so it appears to be the relative path of generate_cubemx.mk. Works on both Windows and Linux (Ubuntu WSL).

Using `$(abspath $(dir $(MAKEFILE_LIST)))` gets an absolute filepath. Is this where the error comes from? Just curious since the current code is quite different with
```
$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
```
If you don't get an error with the abspath version, lets do it. If there is, then I still like your PR suggestion better than the current

BlakeFreer:
I agree that an absolute filepath seems safer, but also there really shouldn't be any error with a relative one. I'll leave the choice up to you, and worst case if someone has an error we will know exactly where to look. 

If you go with relative, update the makefile as required. If you choose absolute, also specify the version requirements in firmware/readme

TylerStAmour:
I ended up testing it some more with newer versions of Make and got the same bug, I guess it has more to do with who's distributing that Windows-ported version of make rather than the actual version number. I also noticed that if you call another makefile with an absolute path (like how I had `@make --file C:/racecar/generate_cubemx.mk --no-print-directory` ) then the issue persists, but if I used a relative path like `./generate_cubemx.mk` then it would be fine when using abspath. 

For compatibility, I think I'll stick to the `$(dir $(MAKEFILE_LIST))` method, I can't see the cubemx generation files being changed much so it hopefully won't cause any issues down the line

BlakeFreer (approved):

Issue #142, opened 2024-08-01
Is Java installed with CubeMX on Linux?


BlakeFreer:
On windows, the STM32CubeMX installer includes a copy of the java runtime environment (JRE) in the `STM32CubeMX/jre` directory. This removes the need for a separate java installation.

Does the Linux installation include the JRE? Currently our documentation assumes not.

Task:
1. Install STM32CubeMX on Linux.
2. See if it includes the JRE
3. If so, update 2 files:
    - `wiki/docs/firmware/dev-setup.md` remove the JRE dependency for Linux
      - since Windows already includes JRE, you can actually remove the whole JRE section from this document
    - `firmware/cmake/generate_cubemx.mk` redirect the `JAVA` variable in the Linux section to use the installed version.

Issue #28, opened 2024-01-13
FW Gen 2 Building Documentation


BlakeFreer:
How to setup the cmake files

- if there is a base project, then most of this should be done already

deepclean if getting "function already defined here" when linking the stm generated files

BlakeFreer:
The documentation in DemoProject should be (mostly) moved out of the project and into a dedicated document

PR #144, opened 2024-08-06
feat: Use the CubeMX Java runtime environment 

BlakeFreer:
The JRE is installed with CubeMX on windows. Using this version removes the need for a separate dependency and guarantees that the JRE version matches Cube.

Edits firmware/cmake/generate_cubemx.mk to use this executable on Windows

I am unsure if Java is packages with cubemx on Mac / Linux so this PR still uses a separate Java installation. See #142


BlakeFreer:
I did not update the firmware/README to show this change. The docs will be updated in #28

TylerStAmour:
Lgtm, haven't tested though

TylerStAmour (approved):

PR #181, opened 2024-09-08
fix: Allocate more stack memory to tasks.

BlakeFreer:
While testing TMS, I came across a strange bug where Update() would hang when the OS was used. Replacing the OS with manual timer & function call avoided the problem. The cause was the large `RawCanMsg[100]` array in `CanBase::ReadQueue()` which exceeded the memory limits of FreeRTOS. This was a very strange error as the memory limit was only slightly exceeded, so calling ReadQueue (even with all internal code commented out) exceeded the memory, but then removing `virtual` from `ReadQueue` removed enough memory to call the function.

To fix this error, I allocated more memory to the FreeRTOS heap and Update task:

1. In CubeMX, go to Middlewares -> FreeRTOS
2. Config -> Total Heap was increased from 15 to 100 kB
3. Tasks & Queues -> Update -> Stack Size was increased from 128 to 4096 words


BlakeFreer:
Other changes:
- Renames LEDs to match TMS board colours
- Removes the unused Log function

PR #184, opened 2024-09-12
Improve CAN generator logging levels (closes #128)

### LINKED ISSUE #128: Improve CAN generator logging levels
# Problem

The cangen python script utilizes logging, which is useful for debugging but is too verbose for regular builds as it pollutes the output with "... shift mask calculated".

To see this output:

1. Clone the racecar repo.
2. Change into the root `racecar/` directory.
3. Create a virtual python environment, activate it, and install the cangen requirements.
```bash
python3.12 -m venv .env
source .env/bin/activate
pip install -r scripts/cangen/requirements.txt
```
> The first two lines are for Linux. On Windows, use `py -3.12 -m venv .env` and `./.env/Scripts/activate`. You may be able to use older versions of python.
4. Run `python scripts/cangen/main.py --project EV5/FrontController`
  > You can run for any project that uses CAN (these projects will have a `config.yaml` file under `firmware/projects/<project_name>`

Output:

```text
INFO    | (can_generator) Parsing DBC files: ['C:\\Formula\\repos\\racecar\\scripts\\cangen\\..\\..\\firmware\\dbcs\\pt.dbc']
INFO    | (can_generator) Successfully added DBC: C:\Formula\repos\racecar\scripts\cangen\..\..\firmware\dbcs\pt.dbc
INFO    | (can_generator) Signal types retrieved
INFO    | (can_generator) Signal types retrieved
INFO    | (can_generator) Filtered messages by node: FC. Num msgs: rx = 4, tx = 2
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
INFO    | (can_generator) Little endian shift amounts calculated.
INFO    | (can_generator) Little endian mask generated.
... and so on
```

There are some useful messages that should be logged, such as where the generated code is placed. Additionally, the developer should be able to decide, at build time, which messages are logged and which are ignored by changing the verbosity level.

## Task

1. Update the logging calls in `main.py` and `can_generator.py` to use more appropriate logging levels.
  > You should also decide which logging level to use for each message. I'd suggest "DEBUG" for most of them, but "INFO" for one-time messages like "Parsing DBC files..." and "Rendered code to (path)". Add other messages as necessary.
2. Add an argument in `parse()` of `main.py` which allows command-line control of the logging level.

### CMake integration

When building a project, `main.py` is executed automatically by CMake (see `firmware/projects/FrontController/CMakeLists.txt`. We would like to change the logging level depending on CMake's internal verbosity level.

See if CMake has a built in verbosity filter, and if so, pass the verbosity level into the cangen script to change the logging verbosity.

Jinwoo-H:
Use flag --log-level INFO or --log-level DEBUG to set verbosity when executing  `python scripts/cangen/main.py` (i.e. `python scripts/cangen/main.py --project EV5/FrontController --log-level INFO`).

`firmware/projects/EV5/LVController/CMakeLists.txt` will pass its internal verbosity level to the Python script when executing it. By default, it is set to STATUS (equivalent to INFO), and also supports VERBOSE (equivalent to DEBUG).


scripts/cangen/main.py
```diff
@@ -45,6 +48,16 @@ def parse():
...
+    
+    # Map CMAKE log levels -> Python log levels
+    log_level_mappings = {
+        "STATUS": "INFO",
+        "INFO": "INFO",
+        "VERBOSE": "DEBUG",
+        "DEBUG": "DEBUG"
+    }
+    # Set log level thrershold to specified verbosity
+    logging.getLogger().setLevel(log_level_mappings[args.level])
```
BlakeFreer, line 60: This line will cause a KeyError if a less verbose cmake level is used (like ERROR or WARNING) since they are not keys in the dictionary. Use `log_level_mappings.get(args.level, "INFO")` which will default to INFO if the key is not found.

scripts/cangen/main.py
```diff
@@ -45,6 +48,16 @@ def parse():
...
     os.chdir(os.path.join(DIR_PROJECTS, project_folder_name))
+    
+    # Map CMAKE log levels -> Python log levels
+    log_level_mappings = {
+        "STATUS": "INFO",
+        "INFO": "INFO",
+        "VERBOSE": "DEBUG",
+        "DEBUG": "DEBUG"
+    }
+    # Set log level thrershold to specified verbosity
```
BlakeFreer, line 59: Typo in "threshold"

scripts/cangen/main.py
BlakeFreer: Nevermind, i see that this case is avoided by the argparse argument choices.

scripts/cangen/main.py
```diff
@@ -33,6 +33,9 @@ def parse():
     parser.add_argument(
         "--project", type=str, required=True, help="Name of the project"
     )
+    parser.add_argument(
+        "--log-level", "-l", dest='level', choices=["STATUS", "INFO", "VERBOSE", "DEBUG"], default="INFO", help="Log verbosity threshold"
```
BlakeFreer, line 37: Since this script is called automatically, I'd rather not have a shorthand `-l` option. Removing this forces the build system developer to write the full `--log-level` in the generator script which is clearer for someone reading it later. 

You can just remove the `"-l",` argument, argparse can have options without short forms.

scripts/cangen/can_generator.py
```diff
@@ -37,7 +37,7 @@ def _assert_valid_dbc(filename: str):
 
 
 def _parse_dbc_files(dbc_files: List[str]) -> Database:
-    logger.info(f"Parsing DBC files: {dbc_files}")
+    logger.debug(f"Parsing DBC files: {dbc_files}")
```
BlakeFreer, line 40: I'd like this to stay "info" since it is a one time message and contains fairly important information

BlakeFreer (changes requested):

BlakeFreer (approved):

PR #187, opened 2024-09-15
Abstract out CAN Gen from CMAkeLists

### LINKED ISSUE #186: Abstract out CAN Gen from CMakeLists
# Problem

Currently, all projects using CAN must repeat the (nearly) identical CMake code:

```cmake
find_package(Python3 COMPONENTS Interpreter)
message(${Python3_EXECUTABLE})

if(NOT ${Python3_FOUND})
	message(FATAL_ERROR "Python 3 executable not found")
endif()

FILE(GLOB_RECURSE CAN_DEPENDENCIES CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/../scripts/cangen/*" "${CMAKE_SOURCE_DIR}/dbcs/*")

list(APPEND CAN_DEPENDENCIES "${CMAKE_CURRENT_SOURCE_DIR}/config.yaml")

cmake_language(GET_MESSAGE_LOG_LEVEL LOG_LEVEL)

add_custom_target(
    generated_can
	COMMAND ${Python3_EXECUTABLE} "${CMAKE_SOURCE_DIR}/../scripts/cangen/main.py" "--project=\"EV5/LVController\"" "--log-level=${LOG_LEVEL}"
	DEPENDS ${CAN_DEPENDENCIES}
	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
	COMMENT Generating CAN code from DBCs
)

# Link ETL which is needed for generated/can/msg_registry.h
add_subdirectory(${CMAKE_SOURCE_DIR}/third-party/etl ${CMAKE_BINARY_DIR}/third-party)
target_link_libraries(main PRIVATE etl)
```

This is not good because it duplicates identical logic in every project and makes it hard to maintain.

Move this code to a new file `firmware/cmake/cangen.cmake`, then use the `include()` method to dynamically include the files contents in the original CMakeLists.txt.

> We already use this structure to centralize the CubeMX generation CMake code. A file is included into each projects' `platforms/stm32f767/cubemx/CMakeLists.txt` file using `include(${CMAKE_SOURCE_DIR}/cmake/build_cubemx.cmake)`. Use this syntax as a guide. `CMAKE_SOURCE_DIR` refers to the directory containing the top-level CMake file, which in our case is `racecar/firmware/`

I say "nearly" equivalent because the `--project=\"<PROJECT_PATH>\"` option must match the current project path (relative to `firmware/projects/`) so that `cangen/main.py` knows where to look for the `config.yaml` file containing CAN configuration.

The CMake file code is run in the context of the file calling `include()`, so you can replace the hard-coded `<PROJECT_PATH>` with a path determined from `CMAKE_CURRENT_SOURCE_DIR` ([reference](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html))

## Task

1. Move the relevant code from `LVController/CMakeLists.txt` to the `firmware/cmake` file specified above, then include the code back into the LV CMakeLists.
    > Use LV because it contains the updated logging from #184 
3. Build the LV Project with
    ```bash
    make PROEJCT=EV5/LVController PLATFORM=cli
    ```
    and make sure that the cangen script still runs.
4. Replace the hard-coded PROJECT value with one computed dynamically from the `CMAKE_CURRENT_SOURCE_DIR` variable.
5. Build LV again to test.
6. Propagate this change to all projects that use CAN (check each `<project>/CMakeLists.txt` file for this code and replace with an `include()`)

Jinwoo-H:
close #186

### DIFF
--- a/firmware/cmake/cangen.cmake
+++ b/firmware/cmake/cangen.cmake
@@ -19,6 +19,8 @@ add_custom_target(
 	COMMENT Generating CAN code from DBCs
 )
 
+add_dependencies(main generated_can)
+
 # Link ETL which is needed for generated/can/msg_registry.h
 add_subdirectory(${CMAKE_SOURCE_DIR}/third-party/etl ${CMAKE_BINARY_DIR}/third-party)
 target_link_libraries(main PRIVATE etl)
--- a/firmware/cmake/cangen.cmake
+++ b/firmware/cmake/cangen.cmake
@@ -1,0 +1,24 @@
+find_package(Python3 COMPONENTS Interpreter)
+message(${Python3_EXECUTABLE})
+
+if(NOT ${Python3_FOUND})
+	message(FATAL_ERROR "Python 3 executable not found")
+endif()
+
+FILE(GLOB_RECURSE CAN_DEPENDENCIES CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/../scripts/cangen/*" "${CMAKE_SOURCE_DIR}/dbcs/*")
+
+list(APPEND CAN_DEPENDENCIES "${CMAKE_CURRENT_SOURCE_DIR}/config.yaml")
+
+cmake_language(GET_MESSAGE_LOG_LEVEL LOG_LEVEL)
+
+add_custom_target(
+    generated_can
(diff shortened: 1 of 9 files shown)


BlakeFreer (changes requested):
I think `add_dependencies(main generated_can)` should also be abstracted out. This is safe since the target name is guaranteed to be `main`. 

I think it makes sense because it ensures that the DBCs actually run when `main` is built, rather than relying on external configuration.

What do you think?

BlakeFreer (approved):

PR #191, opened 2024-09-20
Modify DBC file placement to project specific paths

### LINKED ISSUE #154: DBC File placement
Our CAN DBC files are currently _all_ stored under `firmware/dbcs`. Sharing this folder means it is filled with unrelated DBC files, for example `demo.dbc` for the Demo/Can projects is mixed with the actual EV5 `pt.dbc` and `veh.dbc` files.

```text
firmware/
├── dbcs/
│   ├── demo.dbc
│   ├── pt.dbc
│   └── veh.dbc
└── projects/
    ├── DemoCanReceiver/
    ├── DemoCanTransmitter/
    ├── LVController/
    ├── FrontController/
    └── TMS/
```

It also means that all DBC files must have different names which could become a problem as more versions are created. We cannot maintain the existing `veh.dbc` file and update it for the new season without changing a file name. This makes it difficult to maintain consistent naming across projects.

One solution is placing the DBC files in a project. However, projects (currently) correspond to a single ECU and multiple ECUs need access to the DBCs (since DBC files are the inter-ECU messaging protocol). This solution would incorrectly imply that one project owns the DBC.

Another solution (which ~~will be~~ was enabled by #113) is to organize the _projects_ into folders which the DBC files. This makes a _lot_ of sense given that the DBCs connect multiple ECU projects, so these projects should be combined in a folder. This could look like

```text
firmware/
└── projects/
    ├── DemoCan/
    │   ├── Receiver/
    │   ├── Transmitter/
    │   └── demo.dbc
    └── EV5/
        ├── LVController/
        ├── FrontController/
        ├── TMS/
        ├── pt.dbc
        └── veh.dbc
```

Jinwoo-H:
close #154 

`firmware/dbcs` directory was removed, all files were moved to their respective project directory (i.e. `firmware/projects/EV5`). `firmware/dbcs/VEH_TEST_CAN.dbc` was deleted as it was never referenced in any config.yaml files.


firmware/projects/Demo/io.dbc
BlakeFreer, line 1: This file is used by `EV5/debug/IoCheckoutFc`. It can be placed in that project folder

firmware/projects/EV5/debug/IoCheckoutFc/config.yaml
```diff
@@ -3,4 +3,4 @@ canGen:
   busses:
     - busName: io
       dbcFiles:
-        - "io.dbc"
+        - "../../io.dbc"
```
BlakeFreer, line 6: io.dbc should be moved into the project directory, so reference it with just `"io.dbc"`

firmware/projects/Demo/pedal.dbc
BlakeFreer, line 1: `pedal.dbc` is user by EV5/debug/MotorDebug. Place the dbc in that project

firmware/projects/Demo/demo.dbc
BlakeFreer, line 1: used by Demo/CAN/Foo and /Bar. We want to place the dbcs as far down the hierarchy as possible, so this file can go in the Demo/CAN directory

BlakeFreer (changes requested):

scripts/cangen/main.py
```diff
@@ -73,9 +72,9 @@ def parse():
         # pdb.set_trace()
         bus_name = bus['busName'].capitalize()
         dbc_files = bus['dbcFiles']
-
-        dbc_file_paths = [os.path.join(DIR_DBCS, dbc) for dbc in dbc_files]
-
+        
+        dbc_file_paths = [os.path.join(os.getcwd(), dbc) for dbc in dbc_files]
```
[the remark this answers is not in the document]
BlakeFreer: It would also be good to pass the path through `os.path.normpath` ([docs](https://docs.python.org/3/library/os.path.html#os.path.normpath)) to collapse any relative parent directories. This will make the error messages clearer ex:

```text
Current:
FileNotFoundError: Could not find a file at ...\projects\EV5\debug\IoCheckoutFc\../../io.dbc.
With normpath()
FileNotFoundError: Could not find a file at ...\projects\EV5\io.dbc.
```

BlakeFreer (approved):
Verified Cangen properly generates with all projects.

PR #195, opened 2024-09-30
fixed #175 CAN not working with lowercase

### LINKED ISSUE #175: cangen does not work with some node names
`scripts/cangen` fails to generate code when the ~~bus~~ node name contains lowercase letters.

AndrewI26:
fixed the issue #175, allowing the cangen to work successfully regardless of the capitalization of node names in the .dbc file for a project and the node name in the config.yaml file.


BlakeFreer:
tested and generates with upper, mixed, and lowercase node names

BlakeFreer (approved):
tested and works for upper, lower, and mixed case names

PR #223, opened 2024-10-06
Cleaning CanGen

ManushPatell:


scripts/cangen/main.py
```diff
@@ -68,6 +68,14 @@ def parse():
     bus_list = config["canGen"]["busses"]
     output_path = config["canGen"].get("outputPath", DEFAULT_OUTPUT_DIR)
 
+
+#clear the directory file
+if os.path.exists(output_path):
```
BlakeFreer, line 73: this whole block should be indented to be inside the `'if __name__ == "__main__":` block

scripts/cangen/main.py
```diff
@@ -68,6 +68,14 @@ def parse():
...
+    for filename in os.listdir(output_path):
+        os.remove(os.path.join(output_path, filename))
```
BlakeFreer, line 75: `os.listdir()` also returns the names of subfolders of `output_path`. this would cause `os.remove(...` to fail since it only works on files. We want to delete the entire `generated/can` folder. You could try to recursively search for directories and folders, but a simpler solution is to use [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree) which deletes a folder and its contents.

If you use this function, you can replace the `for` loop with just `shutil.rmtree(output_path)`. This also makes it clearer that we are deleting the entire directory.

scripts/cangen/main.py
```diff
@@ -68,6 +68,14 @@ def parse():
     bus_list = config["canGen"]["busses"]
     output_path = config["canGen"].get("outputPath", DEFAULT_OUTPUT_DIR)
 
+
+#clear the directory file
```
BlakeFreer, line 72: Can you improve this comment? Something about getting rid of old files before making new ones

scripts/cangen/main.py
```diff
@@ -68,6 +68,14 @@ def parse():
     bus_list = config["canGen"]["busses"]
     output_path = config["canGen"].get("outputPath", DEFAULT_OUTPUT_DIR)
 
+
```
BlakeFreer, line 71: remove this extra line, there should be 1 blank line not 2

BlakeFreer (changes requested):

scripts/cangen/main.py
```diff
@@ -68,6 +69,9 @@ def parse():
     bus_list = config["canGen"]["busses"]
     output_path = config["canGen"].get("outputPath", DEFAULT_OUTPUT_DIR)
 
+    # Deletes output path folder and files within, before creating new ones
+    shutil.rmtree(output_path)
```
BlakeFreer, line 73: sorry I should have been more clear, you should still wrap this inside `if os.path.exists(...)`

BlakeFreer (changes requested):

ManushPatell:
Would it throw an error if there was no folder and files to delete. That
seems like the only reason we would need to check if it exists right? Also
no worries, I should have thought of that. I can send another pull request
tomorrow

BlakeFreer:
Yes it would throw an error and we don't want it to crash the program here.

BlakeFreer:
> before I confirm changes, this would be correct then right? It checks if the specified path exists and if so, it will remove it. # Deletes output path folder and files within, before creating new ones if os.path.exists(output_path): shutil.rmtree(output_path)

Yes that is correct.

Next time, you can just make a commit. There's no cost to doing so and it makes it easier to review on GitHub

BlakeFreer (approved):
Looks good, tested on EV5/LVController

Issue #183, opened 2024-09-12
Trailing slash in PROJECT causes wrong binary name


BlakeFreer:
```bash
make PROJECT=EV5/debug/MotorDebug PLATFORM=stm32f767
# produces EV5-debug-MotorDebug.bin
make PROJECT=EV5/debug/MotorDebug/ PLATFORM=stm32f767
# produces EV5-debug-MotorDebug-.bin
```

The built binary names are different due to the trailing slash in the second command. Caused by character substitution in firmware/Makefile. This should not occur; paths should be normalized before generating a name. Alternatively, we could make the binary name come from the final path component (`MotorDebug.bin`) rather than including the whole path relative to `projects/`

This issue was introduced in #156 

---

Since projects are built in separate directories, there is no need to have unique project names. We will use the final component of the project name as the output filename.

For example, the project above will produce `MotorDebug.bin`

# Task

Change how `PROJECT_NAME` is calculated in `firmware/Makefile`

# Verification

`cd` into the `racecar/firmware` directory and build the project with both commands.

```bash
make PROJECT=EV5/debug/MotorDebug platform=stm32f767 clean build
make PROJECT=EV5/debug/MotorDebug/ platform=stm32f767 clean build
```

and ensure that the binary in `firmware/build/EV5/debug/MotorDebug/stm32f767/` is called `MotorDebug.bin` after each command

PR #228, opened 2024-10-11
Update makefile to handle trailing slashes in the project argument in…

Nagatarun25:
Update makefile to adjust for the binary name when a trailing / is specified in the project.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -12,8 +12,7 @@ PLATFORM_DIR := projects/$(PROJECT)/platforms/$(PLATFORM)
 GENERATOR := -G"Unix Makefiles"
 LOG_LEVEL := STATUS
 
-# Name must be a valid filename (i.e. cannot have '/')
-PROJECT_NAME := $(subst /,-,$(PROJECT))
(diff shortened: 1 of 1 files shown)


firmware/Makefile
```diff
@@ -12,8 +12,7 @@ PLATFORM_DIR := projects/$(PROJECT)/platforms/$(PLATFORM)
 GENERATOR := -G"Unix Makefiles"
 LOG_LEVEL := STATUS
 
-# Name must be a valid filename (i.e. cannot have '/')
-PROJECT_NAME := $(subst /,-,$(PROJECT))
+PROJECT_NAME := $(notdir $(patsubst %/,%,$(PROJECT)))
```
BlakeFreer, line 15: Can you add a comment explaining this line, like 'extract project folder name without parent path'

BlakeFreer (changes requested):

firmware/Makefile
Nagatarun25: added the comment

BlakeFreer (approved):

BlakeFreer:
This closes #183

PR #230, opened 2024-10-13
Tarun/interrupt-hooks-issue

Nagatarun25:
Added initial implementation of interrupt hook mechanism for the Linux platform. Set up signal handling to call app-level functions from platform-level interrupts. Tested basic functionality for cross-platform interrupt handling.


firmware/projects/MyBlink/platforms/cli/bindings.cc
```diff
@@ -0,0 +1,42 @@
...
+
+cli::periph::DigitalOutput indicator{"Indicator"};
+
+}  // namespace mcal
+
+//app level function
+bool bindings::toggle = false;
+
+//app level function that takes care of the interupt handler
+void AppLevelFunction(){
```
BlakeFreer, line 20: The point of separating the app and platform layers is to allow our app level functions to be defined once in `main.cc` and work on any platform. Here, you have defined AppLevelFunction in `cli/bindings.cc`, which means you would have to re-implement it in `stm32f767/bindings.cc` to compile on that platform.

This function should be moved to `main`, then you can `extern` reference it in bindings to attach it to the signal handler

BlakeFreer (changes requested):

firmware/projects/Demo/Interrupts/README.md
```diff
@@ -0,0 +1,21 @@
+# Interrupt Handling Project
+
+The Interrupt Handling project demonstrates how to implement interrupt handling in C++. It is designed to work on CLI platforms and provides an example of managing asynchronous events through signal interrupts.
+
+### Key Components:
+- **Bindings** (this module): Manages the interrupt logic and state control.
```
BlakeFreer, line 6: why did you write "this module"? I don't agree with its use here

firmware/projects/Demo/Interrupts/README.md
```diff
@@ -0,0 +1,21 @@
+# Interrupt Handling Project
+
+The Interrupt Handling project demonstrates how to implement interrupt handling in C++. It is designed to work on CLI platforms and provides an example of managing asynchronous events through signal interrupts.
+
+### Key Components:
```
BlakeFreer, line 5: Markdown headers should only increment by one level at a time, so this header should be `##` not `###`. Same for the following headers.

BlakeFreer (changes requested):
Move the rest of the MyBlink files to Demo/Interrupt

BlakeFreer (changes requested):
All of the cmake files re missing from this project so I cannot build it

firmware/projects/Demo/Interrupts/platforms/cli/bindings.cc
BlakeFreer, line 1: Since we are getting rid of `setup_interrupt`, move `signal` to `Initialize()`. I'd also like you to define a new function `ToggleInterruptHandler()`, outside of the namespace, which internally just calls `AppLevelFunction`, then register that handler in `signal` instead. This seems roundabout, but is more generic and will work for other platforms

BlakeFreer (changes requested):

.pre-commit-config.yaml
BlakeFreer, line 1: What's this change for?

firmware/projects/Demo/Interrupts/main.cc
```diff
@@ -0,0 +1,36 @@
+#include "bindings.h"
+#include <iostream> 
+
+
+//toggle variable
+volatile bool bindings::toggle = false;
+
+//app level function that takes care of the interupt handler
+namespace bindings {  // Define AppLevelFunction within the bindings namespace
```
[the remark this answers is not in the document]
BlakeFreer: Still not resolved. App function should not be in the bindings namespace or header file

firmware/projects/Demo/Interrupts/platforms/cli/bindings.cc
```diff
@@ -0,0 +1,32 @@
...
+
+namespace bindings {
+
+    shared::periph::DigitalOutput& indicator = mcal::indicator;
+
+    void DelayMS(unsigned int ms) {
+        usleep(ms * 1000);
+    }
+
+    // interupt function that uses the signal function for the cli/linux
```
BlakeFreer, line 24: What's this comment for?

firmware/projects/Demo/Interrupts/bindings.h
```diff
@@ -0,0 +1,14 @@
...
+#include "shared/periph/gpio.h"
+
+namespace bindings {
+
+    extern shared::periph::DigitalOutput& indicator;
+    extern void DelayMS(unsigned int ms);
+    extern void Initialize();
+    extern void AppLevelFunction();
+
+    }  // namespace bindings
```
BlakeFreer, line 12: ~~See comment about code formatting~~

Since you've already made unformatted commits, this code will not be fixed by that. You should run these commands from inside the Demo/Interrupts folder

pip install clang-format
clang-format *

That will fix all files. Commit them, and the pre commit action should succeed

BlakeFreer (changes requested):

firmware/projects/Demo/Interrupts/platforms/cli/bindings.cc
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
```
BlakeFreer, line 1: What needs this include? Remove it if nothing uses it

BlakeFreer (changes requested):

firmware/projects/Demo/Interrupts/bindings.h
Nagatarun25: I ran these commands within the Demo/interrupts folder. Let me know if this fixed the files

firmware/projects/Demo/Interrupts/main.cc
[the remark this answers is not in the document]
Nagatarun25: I updated the volatile variable toggle and removed the app function from the bindings namespace and header file.

Issue #164, opened 2024-08-26
Split PWMOutput to PWMOutput and AnalogOutput


BlakeFreer:
# PWM Output

Timeline: End of October

## Background Reading

- <https://learn.sparkfun.com/tutorials/pulse-width-modulation/all>

## Description

PWM is a microcontroller peripheral for outputting an Analog voltage level, but it can also be used when digital signal of a specific frequency and duty cycle is required. These are fundamentally different purposes, but the `PWMOutput` peripheral in `racecar/` is responsible for both roles. In this issue, you will separate the functionality into 2 classes.

## Current State

See `firmware/shared/periph/pwm.h`

There are 4 methods on the PWM class:

```c++
class PWMOutput : public util::Peripheral {
public:
    virtual void Start() = 0;
    virtual void Stop() = 0;
    virtual void SetDutyCycle(float duty_cycle) = 0;
    virtual float GetDutyCycle() = 0;
};
```

`Start` and `Stop` control whether the output is active, while `SetDutyCycle` changes the duty cycle and `GetDutyCycle` returns it. This should be broken into 2 different classes

### Class 1: `PWMOutput`

This should be identical to the current PWMOutput but with an added `SetFrequency` and `GetFrequency` method which operates on floating-point frequencies measured in Hertz.

### Class 2: `AnalogOutput`

This class is much different from the existing PWMOutput class. It has only 1 method called `Set` which takes a floating point voltage argument.

### Stage 1: Prototype

1. Create a new header file `firmware/shared/periph/analog_output.h` to define the AnalogOutput abstract class.
2. Edit the existing `shared/.../pwm.h` class as described about
3. Implement the classes for the CLI platform.
4. Make a Demo project which uses your peripherals on the CLI platform.
5. Push to a branch on GitHub

### Stage 2: Implementation

After discussing your implementation, we will apply it to the other platforms and update our existing projects.

For the stm32f767 platform, both AnalogOutput and PWMOutput will use a PWM peripheral under-the-hood. This is fine: the App-Level doesn't care how the peripherals are implemented, only that they have a specific form.

BlakeFreer:
Yes there should be some clamping and it will happen at the platform level. You do not need to clamp on the `cli` but will need to clamp when settings the duty cycle in the `stm32` peripheral. The stm runs at 3.3, not 5V

BlakeFreer:
closed by #231

PR #231, opened 2024-10-14
Adds PWMOutput and AnalogOutput to stm32f767 mcal

floralandfaded:


firmware/mcal/cli/periph/analog_output.h
BlakeFreer, line 1: Can you install the `pre-commit` action from <https://macformula.github.io/racecar/firmware/dev-setup/#pre-commit-setup>? It will re-format your code before it is committed. Unformatted code is blocked from `main` so this will help you avoid the issue before it happens

firmware/mcal/stm32f767/periph/analog_output.h
```diff
@@ -0,0 +1,33 @@
...
+
+class AnalogOutput : public shared::periph::AnalogOutput {
+public:
+    AnalogOutput(TIM_HandleTypeDef* htim, uint32_t channel)
+        : htim_(htim), channel_(channel) {}
+
+    void SetVoltage (float voltage) override {
+        voltage_ = shared::util::Clamper<float>::Evaluate(voltage, 0, 3.3);
+        uint32_t pulse = (voltage_ * float(__HAL_TIM_GetAutoreload(htim_))) /
+                 (3.3 * 100.0f);
```
BlakeFreer, line 22: replace `3.3` with `3.3f`

BlakeFreer: Even though we clamp the voltage, I'm worried that a rounding error could cause the value on the right of `pulse =` to be greater than the TIM Autoreload max, and I think we would get undesired behaviour.

For example, if Autoreload is 10 bit [0-1023] and at voltage=3.3, we could have `pulse=1024`. When this value is placed in the stm register, the last 10 bits are all 0, so the duty cycle might actually be 0 instead of 100%.

Can you find a way to avoid this issue? Can we clamp `pulse` after the conversion? What would it be clamped to?

firmware/mcal/stm32f767/periph/pwm.h
```diff
@@ -60,10 +60,22 @@ class PWMOutput : public shared::periph::PWMOutput {
         return float(pulse) / float(period) * 100.0f;
     }
 
+    void SetFrequency(float frequency) override {
+        frequency_ = std::max((float) 0, frequency);
```
BlakeFreer, line 64: instead of casting `0` to float, just use `0.0f`

BlakeFreer: In fact, clamping to `0` will cause a divide by zero error, so our min value should be something like 0.5 Hz. Can you find the stm's lowest possible PWM frequency?

firmware/projects/Demo/AnalogPWM/main.cc
```diff
@@ -0,0 +1,57 @@
...
+    pwm.GetFrequency();
+    DelayMS(1000);
+
+    pwm.Stop();
+}
+
+/**
+ * @brief Tests the AnalogOutput class using various voltages.
+ */
+void DemoAnalog() {
```
BlakeFreer, line 39: This method doesn't do much on STM since it cannot output 50V. Can you change the values to be within 0 and 3.3 V, maybe add a few more values

BlakeFreer (changes requested):

firmware/mcal/stm32f767/periph/pwm.h
samparent97: rather than remapping to 0 to any value, I think we should make it clear an error has occured. Since we do not have good error support, we should maybe do nothing for now or turn the pwm off completely.

BlakeFreer: This is the plan once errors are ready. If I set the freq too low, I would expect the PWM to still run but at a different freq. Turning it off completely would be harder to debug than operating at the wrong frequency

firmware/mcal/stm32f767/periph/analog_output.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace mcal::stm32f767::periph {
+
+class AnalogOutput : public shared::periph::AnalogOutput {
+public:
+    AnalogOutput(TIM_HandleTypeDef* htim, uint32_t channel)
+        : htim_(htim), channel_(channel) {}
+
+    void SetVoltage (float voltage) override {
+        voltage_ = shared::util::Clamper<float>::Evaluate(voltage, 0, 3.3f);
```
BlakeFreer, line 20: This first clamp is redundant now that we clamp the pulse

firmware/mcal/stm32f767/periph/analog_output.h
```diff
@@ -0,0 +1,36 @@
...
+        uint32_t pulse_ =
+            shared::util::Clamper<float>::Evaluate(pulse, 0, maxPulse);
+
+        __HAL_TIM_SetCompare(htim_, channel_, pulse_);
+    }
+
+private:
+    TIM_HandleTypeDef* htim_;
+    uint32_t channel_;
+    float voltage_;
```
BlakeFreer, line 33: We don't need to remember the voltage, so this variable should be removed. If we ever add a GetVoltage method then we will recalculate the voltage from the register values rather than by recalling a variable.

firmware/mcal/stm32f767/periph/analog_output.h
```diff
@@ -0,0 +1,36 @@
...
+namespace mcal::stm32f767::periph {
+
+class AnalogOutput : public shared::periph::AnalogOutput {
+public:
+    AnalogOutput(TIM_HandleTypeDef* htim, uint32_t channel)
+        : htim_(htim), channel_(channel) {}
+
+    void SetVoltage (float voltage) override {
+        voltage_ = shared::util::Clamper<float>::Evaluate(voltage, 0, 3.3f);
+        uint32_t pulse = (voltage_ * float(__HAL_TIM_GetAutoreload(htim_))) /
```
BlakeFreer, line 21: why do we have a factor of `1/100.0`? I don't think this is correct. If `Autoreload` = 1000, then I need my pulse to be 1000 to set voltage = 3.3V, but this calculation gives 10. Similarly `maxPulse` should be the maximum register value which is a raw integer, not a float

BlakeFreer: You should first calculate `uint32_t maxPulse`, then `pulse = Clamp(voltage / 3.3 * maxPulse, 0, maxPulse)`

firmware/mcal/stm32f767/periph/pwm.h
```diff
@@ -60,10 +60,22 @@ class PWMOutput : public shared::periph::PWMOutput {
         return float(pulse) / float(period) * 100.0f;
     }
 
+    void SetFrequency(float frequency) override {
+        frequency_ = std::max(0.000015259022f, frequency);
```
BlakeFreer, line 64: make this minimum a named constant to express why we have it and where it comes from `kMinimumFrequency = 0.00...`

firmware/mcal/stm32f767/periph/pwm.h
```diff
@@ -60,10 +60,22 @@ class PWMOutput : public shared::periph::PWMOutput {
...
+        uint32_t autoreload = 1/frequency_;
```
BlakeFreer, line 65: This expression assumes that the PWM timer is counting at 1Hz, but in reality it is significantly faster and variable. You need to get the timer frequency to correctly perform this calculation.

Consider writing a private `GetTimerFrequency()` method

firmware/mcal/stm32f767/periph/pwm.h
```diff
@@ -60,10 +60,22 @@ class PWMOutput : public shared::periph::PWMOutput {
...
+
+        __HAL_TIM_SetAutoreload(htim, autoreload);
+    }
+
+    float GetFrequency() override {
+        return frequency_;
```
BlakeFreer, line 71: This should calculate the frequency from the timer reload and timer speed (see GetTimerFrequency comment above)

firmware/mcal/stm32f767/periph/pwm.h
```diff
@@ -60,10 +60,22 @@ class PWMOutput : public shared::periph::PWMOutput {
...
+    }
+
 private:
     TIM_HandleTypeDef* htim_;
     uint32_t channel_;
     float duty_cycle_ = 0;
+    float frequency_;
```
BlakeFreer, line 78: we don't need to store duty_cycle or frequency

BlakeFreer (changes requested):

firmware/mcal/stm32f767/periph/pwm.h
BlakeFreer, line 1: Forgot to comment earlier, but we should not store `duty_cycle_` in a variable. GetDutyCycle should recalculate it from the physical registers

BlakeFreer (changes requested):

firmware/mcal/stm32f767/periph/pwm.h
BlakeFreer: I realize that this is my old code and I justified the decision to store `duty_cycle_` and return it inside `GetDutyCycle`. This was a hacky fix and can provide incorrect information to the caller. Please remove you can ignore the comment above and I will create a separate issue to fix this.

BlakeFreer: See #235

BlakeFreer (approved):

BlakeFreer:
Closes #164

PR #233, opened 2024-10-16
Turned cangen into a module, and refactored cangen

### LINKED ISSUE #204: Make CANgen a Python Package
# CANgen Python Package

Timeline: Mid-late October

## Background

CANgen is a python program which parses CAN `.dbc` files and generates C++ code for our firmware projects. It is executed automatically as CMake compiles a project. The source code is under `racecar/scripts/cangen`.

## Problem

CANgen is currently "just" a set of python scripts which get executed. There is no real 
structure to the package, and installing the dependencies is somewhat arbitrary.

## Tasks

1. Refactor CANgen into a proper Python package using a packaging tool. I highly recommend [Poetry](https://python-poetry.org/). The package dependencies can be specified in `pyproject.toml`, removing the need for `requirements.txt`.

2. Set up a Python linter and formatter to keep your code tidy. Use [`ruff`](https://docs.astral.sh/ruff/).

## Next Steps

#211

### LINKED ISSUE #211: CANgen Entry Point
# CANgen Entry Point

Define `main` as an entry point / script (see <https://python-poetry.org/docs/pyproject/#scripts>). A script is a function that can be called directly. This means we could install your new `cangen` package, then run

        cangen --project=EV5/TMS

    instead of

        python ../scripts/cangen/main.py --project=EV5/TMS

    This has obvious benefits for cross-platform portability and making the build system more modular.

## Next Steps

#212

### LINKED ISSUE #212: CANgen Refactor
# CANgen Refactor

Clean up the python scripts. There is not a clear boundary between `main.py` and `can_generator.py`. Global variables are everywhere. Utility functions (like `_camel_to_snake` and `_decimal_to_hex`) should be moved into their own module, not mixed with all of the CAN code. Etc.

### Important

When refactoring the python scripts, make sure that they generate the _exact_ same C++ code as they do now. You can check this by generating code with your updates, then checking out a commit on `main` to generate with the old scripts.

AndrewI26:
Refactored cangen by adding config and bus objects. Turned cangen into a package with poetry. 

 Fixes #211, fixes #204, fixes #212


scripts/cangen/README.md
BlakeFreer, line 1: Add some details about this package, such as

- What the package does (generate C++ code for CAN messages from a DBC file)
    - What this tool replaces (manually writing C++ code) and why (to improve speed and reduce errors)
- How to manually call use it (`pip install scripts/cangen`, `cangen firmware/projects/...`)
- Where it is called in the automated build process.
- Anything else that someone might need to know to use and understand this project

The title does not need to include "poetry" or "python". Those are implementation details that shouldn't be highlighted. Recall that poetry is a tool that only you (the developer) needs to have.

You should get in the habit of looking at well-known repositories and seeing how they organize their READMEs. Since our tool replaces `cantools`, you might find inspiration from its readme <https://github.com/cantools/cantools/tree/master>

scripts/cangen/cangen/Bus.py
BlakeFreer, line 1: python files should be lowercase, so rename to `bus.py` and fix the imports in other files. Refer to this section of PEP8 (the python style guide) <https://peps.python.org/pep-0008/#package-and-module-names>

Since windows filename are case-INsensitive, you might not be able to directly rename it, so go `Bus.py -> temp.py -> bus.py`

scripts/cangen/cangen/Config.py
BlakeFreer, line 1: see comment on Bus.py, rename to `config.py` (lowercase)

scripts/cangen/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[tool.poetry]
+name = "cangen"
+version = "0.1.0"
+description = "Converts .dbc files into C++ files for a CAN bus system."
+authors = ["Your Name <you@example.com>"]
```
BlakeFreer, line 5: update this with your info

scripts/cangen/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.10"
```
BlakeFreer, line 9: where did you come up with these dependencies? I feel like most of them are sub-dependencies of Jinja2 and cantools. We should only list the dependencies that we directly use and let `pip` figure out the sub-dependencies.

To simplify this list, remove all dependencies (besides `python`) and `pip install` this project into a fresh virtual environment. Then run `cangen` and see which packages are missing. Start re-adding the packages, starting with `Jinja2` and `cantools`, until you don't have any errors.

You should use `poetry add package_name` to add the packages instead of manually writing them into the toml file. See <https://python-poetry.org/docs/cli/#add>

scripts/cangen/cangen/util.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def _camel_to_snake(text):
+    """Converts UpperCamelCase to snake_case."""
+
+    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", text)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
+
+
+def _decimal_to_hex(decimal_value):
```
BlakeFreer, line 11: We should get rid of this function and just use `hex` directly. It's an unnecessary abstraction (ik you didn't write this, but we should clean it out)

scripts/cangen/cangen/util.py
BlakeFreer, line 1: After removing the decimal to hex function, this module will only contain the camel to snake function. To simplify the layout, let's move the function into the module where it is used

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+import argparse
+import logging
+
+from .generate_can_from_dbc import generate_can_from_dbc
+
+logging.basicConfig(level="INFO", format="%(levelname)-8s| (%(name)s) %(message)s")
+
+CONFIG_FILE_NAME = "config.yaml"
```
BlakeFreer, line 13: unused, remove

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+        "--log-level",
+        dest="level",
+        choices=["STATUS", "INFO", "VERBOSE", "DEBUG"],
+        default="INFO",
+        help="Log verbosity threshold",
+    )
+
+    # If parsing fails (ex incorrect or no arguments provided) then this exits with
+    # code 2.
+    return parser.parse_args()
```
BlakeFreer, line 29: Since this program always requires CLI arguments, we can show the help screen if `cangen` is called with no arguments. Replace `return parser.parse_args()` with 

```python
return parser.parse_args(None if sys.argv[1:] else ["-h"])
```

You will need to import `sys` for this to work. Try to figure out what this expression evaluates to when different (or no) CLI args are passed, and why it shows the help screen when there aren't any CLI args

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # Map CMAKE log levels -> Python log levels
+    log_level_mappings = {
+        "STATUS": "INFO",
+        "INFO": "INFO",
+        "VERBOSE": "DEBUG",
+        "DEBUG": "DEBUG",
+    }
+    # Set log level threshold to specified verbosity
+    logging.getLogger().setLevel(log_level_mappings[args.level])
```
BlakeFreer, line 43: This modifies the global python logger. We should only set the logging level for our package. Define a logger for this module using the exact line of code from the top of `can_generator.py` and replace `logging.getLogger().setLevel` with `logger.setLevel`

scripts/cangen/cangen/generate_can_from_dbc.py
BlakeFreer, line 1: There's no need for this to be it's own file. It will be easier for someone to follow if it is in the same file as `generate_code`. Move these contents into can_generator.py and delete this file

scripts/cangen/cangen/Config.py
```diff
@@ -0,0 +1,18 @@
+import yaml
+from .Bus import Bus
+
+
+class Config:
+    OUTPUT_PATH = "generated/can"
+
+    def __init__(self, config_file_name: str) -> None:
```
BlakeFreer, line 8: constructors aren't usually annotated with `-> None`

scripts/cangen/cangen/can_generator.py
```diff
@@ -22,43 +24,47 @@
...
+MSG_REGISTRY_TEMPLATE_FILENAME = "msg_registry.h.jinja2"
 
-    logger.debug(f"{filename} is a valid dbc file.")
+DIR_THIS_FILE = os.path.abspath(os.path.dirname(__file__))
+DIR_TEMPLATES = os.path.join(DIR_THIS_FILE, "templates")
+CAN_MESSAGES_TEMPLATE_PATH = os.path.join(DIR_TEMPLATES, CAN_MESSAGES_TEMPLATE_FILENAME)
+MSG_REGISTRY_TEMPLATE_PATH = os.path.join(DIR_TEMPLATES, MSG_REGISTRY_TEMPLATE_FILENAME)
 
 
-def _parse_dbc_files(dbc_files: List[str]) -> Database:
```
BlakeFreer, line 39: this function looks good

scripts/cangen/cangen/Bus.py
BlakeFreer: Actually, python modules don't separate small classes into individual files (this is very different from C++ and Java). This file's contents should be moved into `config.py`

scripts/cangen/cangen/Config.py
```diff
@@ -0,0 +1,18 @@
+import yaml
+from .Bus import Bus
+
+
+class Config:
+    OUTPUT_PATH = "generated/can"
```
BlakeFreer, line 6: How is this value being used?

BlakeFreer (changes requested):

firmware/projects/EV5/debug/FrontControllerSimple/config.yaml
```diff
@@ -2,8 +2,6 @@ canGen:
   ourNode: fc
   busses:
     - busName: veh
-      dbcFiles:
-        - "../../veh.dbc"
+      dbcFile: "../../veh.dbc"
```
samparent97, line 5: @BlakeFreer wouldn't it be better to give paths starting from racecar instead of using relative paths. 

For example this would become
```yaml
    dbcFile: "firmware/projects/EV5/veh.dbc"
```

I understand for simple projects this becomes verbose, but for projects that start to have more nested folder structures like this one, it seems pretty messy to have it this way.

scripts/cangen/README.md
```diff
@@ -0,0 +1,36 @@
+# Cangen
+Cangen is a cli package used to generate C++/C files from a corresponding dbc file. This tool automates the can generation process and ensures can generation is consistent between all projects. 
+
+### Installation
+First install the cangen package through the terminal:
+
+```
+pip install racecar/scripts/cangen
```
samparent97, line 8: Is there a path I need to run this from?

firmware/projects/EV5/debug/FrontControllerSimple/config.yaml
BlakeFreer: Projects should be nested according to their logical grouping with other projects. CAN buses are defined for such a group of projects, so the nesting will not become too deep.

Using an absolute path like `firmware/.../veh.dbc` couples the config.yaml to the folder structure beyond the project group. If I rename or move the folder holding a group of projects and associated DBCs, then the config.yaml will need to be updated if it uses absolute paths, whereas it wouldn't using these relative paths.

Also `firmware/projects/EV5/veh.dbc` is much  more verbose than `../../veh.dbc`

scripts/cangen/README.md
BlakeFreer: This command installs into the current python virtual environment (or global environment if no venv is activated) and it just needs the path to the `cangen` folder. You could run `pip install cangen` from within `racecar/scripts` or even `pip install ../../scripts/cangen` from inside `racecar/firmware/mcal`.

We should update the docs to install `scripts/cangen` and tell the user to run it from the `racecar/` directory

AndrewI26: I updated the readme and recommitted the code.

scripts/cangen/README.md
```diff
@@ -0,0 +1,41 @@
+# Cangen
+Cangen is a cli package used to generate C++/C files from a corresponding dbc file. This tool automates the can generation process and ensures can generation is consistent between all projects.
```
BlakeFreer, line 2: CLI, DBC, CAN should all be capitalized. Since we generate C++, you can remove the "/C"

BlakeFreer, line 2: Remove the start of this sentence. We do not need to state that cangen is a package; that is obvious. The reader is interested in what cangen does. This sentence should read like

"Generate C++ code from a CAN DBC file. Ensures repeatable generation between all projects."

scripts/cangen/README.md
```diff
@@ -0,0 +1,41 @@
+# Cangen
+Cangen is a cli package used to generate C++/C files from a corresponding dbc file. This tool automates the can generation process and ensures can generation is consistent between all projects. 
+
+### Installation
+First install the cangen package through the terminal:
+
+```
+pip install racecar/scripts/cangen
+```
+The directory where you run this command does not matter, as long as the path specified leads to the directory containing the pyproject.toml for cangen.
```
BlakeFreer, line 10: I didn't mean for my reply to Sam to be included in the README, that was for his knowledge. We should be clear here and provide just a single installation command.

Replace the "Installation" section with

````markdown
## Installation

From the `racecar/` directory, run

```bash
pip install scripts/cangen
```
````

Also notice that I used only 2 `##` before the header instead of 3 `###`. Markdown headers should only increment by one level at a time. You should also fix this in the Usage section

File: scripts/cangen/README.md [hunk 14]
```diff
@@ -0,0 +1,41 @@
...
+For example, if you are in the racecar directory already, then you could run 
+```
```
BlakeFreer, line 12: There's a few markdown formatting conventions, like newlines, that can be automatically applied using the `markdownlint` vscode extension. Can you install it and run `> Format Document With` -> `markdownlint`? You can even set up your editor to "Format on Save" to automate this process. Like ruff for python, it keep our documentation consistent

scripts/cangen/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+pip install racecar/scripts/cangen
+```
+Make sure to specify the relative path to the cangen directory. 
+
+### Example usage
+To generate code for a project, like FrontController:
+```
+cangen racecar/firmware/projects/EV5/FrontController
```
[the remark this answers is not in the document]
BlakeFreer: Please update usage section as above

scripts/cangen/README.md
```diff
@@ -0,0 +1,41 @@
...
+pip install scripts/cangen
+```
+
+
+### Example usage
+To generate code for a project, like FrontController:
+```
+cangen racecar/firmware/projects/EV5/FrontController
+```
+The file path should be the relative path to the project you are looking to generate code for.
```
BlakeFreer, line 22: Remove this line, it is implicit in the previous instuctions

scripts/cangen/README.md
```diff
@@ -0,0 +1,41 @@
...
+This command will generate code in a generated/can folder inside the selected projects directory.
```
BlakeFreer, line 23: generated/can should be enclosed with backticks, i.e. replace

```markdown
generated/can/
```

with

```markdown
`generated/can`
```

BlakeFreer, line 23: ... in a \`generated/can/\` subfolder of the project.

scripts/cangen/cangen/can_generator.py
```diff
@@ -22,43 +24,47 @@
 MSG_REGISTRY_FILE_NAME = "_msg_registry.h"
 CAN_MESSAGES_FILE_NAME = "_can_messages.h"
 
+OUTPUT_DIR = "generated/can"
```
BlakeFreer, line 27: This should be moved to config.py and used as a default argument to config.pop('outputPath')

scripts/cangen/cangen/can_generator.py
```diff
@@ -266,19 +252,33 @@ def generate_code(
         "bus_name": bus_name,
     }
 
-    # Replace these lines with your Jinja2 template logic
     logger.debug("Generating code for can messages")
     _generate_from_jinja2_template(
-        can_messages_template_path,
-        os.path.join(output_dir, bus_name.lower() + CAN_MESSAGES_FILE_NAME),
+        CAN_MESSAGES_TEMPLATE_PATH,
+        os.path.join(OUTPUT_DIR, bus_name.lower() + CAN_MESSAGES_FILE_NAME),
```
BlakeFreer, line 258: should be using `config.output_path` not OUTPUT_DIR

scripts/cangen/cangen/can_generator.py
```diff
@@ -266,19 +252,33 @@ def generate_code(
...
     )
 
     logger.info("Code generation complete")
+
+
+CONFIG_FILE_NAME = "config.yaml"
+
+
+def generate_can_from_dbc(project_folder_name: str):
+    # Change directory to the project folder
```
BlakeFreer, line 276: remove all 3 of these comments. the code is very easy to understand so comments only slow the reader

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+    args = parse()
+
+    # Map CMAKE log levels -> Python log levels
+    log_level_mappings = {
+        "STATUS": "INFO",
+        "INFO": "INFO",
+        "VERBOSE": "DEBUG",
+        "DEBUG": "DEBUG",
+    }
+    # Set log level threshold to specified verbosity
```
BlakeFreer, line 42: don't need this comment

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+        "--log-level",
+        dest="level",
+        choices=["STATUS", "INFO", "VERBOSE", "DEBUG"],
+        default="INFO",
+        help="Log verbosity threshold",
+    )
+
+    # If parsing fails (ex incorrect or no arguments provided) then this exits with
+    # code 2.
+    return parser.parse_args(None if sys.argv[1:] else ["-h"])
```
BlakeFreer, line 29: ik I asked you to add the `None if ...` but this violates the comment above which states that the program will exit with code 2 if parsing fails.

Since this tool is primarily used by other scripts, not by humans, I'd prefer the error code instead of the help message. With this setup, there is no way to distinguish success from the failed run where no args are passed.

scripts/cangen/cangen/main.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def main():
+    args = parse()
+
+    # Map CMAKE log levels -> Python log levels
```
BlakeFreer, line 35: we can make the code self-documenting by renaming the variable from "log_level_mappings" to "cmake_to_python_level" then removing the comment

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
+import yaml
+
+
+class Bus:
+    def __init__(self, bus: dict):
+        self.bus = bus
```
BlakeFreer, line 6: don't store `self.bus`, we want it to be empty at the end of `__init__`

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
...
+
+        self.dbc_file_path: str = self.bus.pop("dbcFile")
+        self.bus_name: str = self.bus.pop("busName").capitalize()
+
+        if not self.has_bus_been_consumed():
```
BlakeFreer, line 11: don't need an extra method here, just call `if bus` directly, or `if bus.keys()`. Same for the Config class

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
+import yaml
+
+
+class Bus:
+    def __init__(self, bus: dict):
+        self.bus = bus
+
+        self.dbc_file_path: str = self.bus.pop("dbcFile")
```
BlakeFreer, line 8: good

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
...
+        self.bus_name: str = self.bus.pop("busName").capitalize()
+
+        if not self.has_bus_been_consumed():
+            print(self.bus)
```
BlakeFreer, line 12: You can pass a message to `raise ValueError(msg_goes_here)`. This message will be printed to the console when the error is raised (try it out). This is better practice than using `print`.

Replace the contents of the `if` block with

```python
raise ValueError(f"Unexpected bus keys {'.'.join(bus.keys())}")
```

Do the same thing for Config, but replace "bus" with "config"

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class Config:
+    @staticmethod
+    def load_yaml(config_file_name: str) -> dict:
+        with open(config_file_name, "r") as file:
+            config = yaml.safe_load(file)
+        return config.pop("canGen")
+
+    def __init__(self, config_file_name: str):
```
BlakeFreer, line 30: `__init__` should take a dictionary, not a filename

```python
def __init__(self config: dict):
   ...
```

init should not handle the file at all. Rather, the `load_yaml` method show be replaced with a `from_yaml` method with the form

```python
@staticmethod
def from_yaml(file_name: str) -> Config:
    with open(filename ...
        config = ...
    return Config(config.pop("canGen"))
```

This is a big difference. It means that we can create Configs from any dictionary, but provide a convenient method for creating one from known yaml format. You code in `can_generator.py` will now be `config = Config.from_yaml(...`

Other details
- Since we are in the Config class, it is redundant to have "config_" in the filename variable
- You will need `from __future__ import annotations` to use the `-> Config` signature

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
...
+        self.config = Config.load_yaml(config_file_name)
+        self.node = self.config.pop("ourNode")
+
```
BlakeFreer, line 33: still need to get `self.output_path`

BlakeFreer (changes requested):

scripts/cangen/cangen/config.py
```diff
@@ -0,0 +1,47 @@
...
+        # Create a bus object for all busses in the config file
+        self.busses = [Bus(bus) for bus in self.config.pop("busses")]
+
+        # Checks to make sure config object is empty, all fields have been popped
```
BlakeFreer, line 37: don't need any comments in __init__, it's all self-explanatory

BlakeFreer (changes requested):

scripts/cangen/cangen/can_generator.py
```diff
@@ -226,6 +224,7 @@ def generate_code(bus: Bus, config: Config):
     dbc_file = bus.dbc_file_path
     our_node = config.node
     bus_name = bus.bus_name
+    output_dir = Config.output_dir
```
BlakeFreer, line 227: few things here:

1. `Config` is a class. You should be accessing the `config` variable.
2. The variable is not called `.output_dir` in the class definition
3. You don't need to make a variable here. Just directly use the class member in the `os.path.join` statements below

BlakeFreer (changes requested):
Make sure you run the code before pushing, your current code produces an error on my comment

PR #241, opened 2024-10-23
Equation for Can Traffic

ManushPatell:
Equation to determine the traffic load on our CAN Busses


docs/docs/can-traffic.md
BlakeFreer, line 1: I'd like you to move this file to `docs/docs/firmware/can-traffic.md`. Use `git mv`.

This file will not appear on our website since it is not included in the `nav` tree. Add it to the `nav:` section of docs/mkdocs.yml under "Firmware". Then when you run `mkdocs serve` (as described [here](https://macformula.github.io/racecar/tutorials/site-dev/)) you will see it added to the sidebar under `Firmware`

docs/docs/can-traffic.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. Total Data Per Cycle
+
+Total data per cycle includes all the bits transmitted in one CAN message. This includes not only the payload but overhead bits like start of frame, CRC, acknowledgement, control field etc. 
+
+
+
+## The bus load is determined by: 
+
+![Bus Load Formula](https://latex.codecogs.com/png.latex?\text{Bus%20Load%20(\%)}%20=%20\frac{\text{Number%20of%20Bits%20per%20Frame}%20\times%20\text{Number%20of%20Frames%20per%20Second}}{\text{CAN%20Speed}}%20\times%20100)
```
BlakeFreer, line 25: Great use of LaTeX, but we should not rely on this external codecogs renderer. Material for mkdocs has a tutorial for using LaTeX directly.

<https://squidfunk.github.io/mkdocs-material/reference/math/?h=mathja#katex>

docs/docs/can-traffic.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+In order to calculate the CAN load, we must declare a few variables and their importance within the CANbus protocol. 
+
+1. Baud: (kbaud)
+
+The baud rate is the rate at which data is transfered in a communication channel (like CAN). For example, 9600 Baud rate is 9600 bits per second
+
+2. Frequency (Hz)
```
BlakeFreer, line 13: Frequency should be divided into frequency per message since some message are transmitted more frequently.

For example, I might give you

| message | frequency (Hz) |
| :---|:---|
| BatteryStatus | 10 |
| DebugCode | 1 |
| MotorControl | 500 |

You calculation should calculate the total bits per second from this table. Are all messages likely to have the same number of bits?

docs/docs/can-traffic.md
```diff
@@ -0,0 +1,38 @@
+---
```
BlakeFreer, line 1: You can make a title in plain markdown using

```markdown
# CAN Traffic Calculation
```

instead of this yml block

docs/docs/can-traffic.md
```diff
@@ -0,0 +1,38 @@
...
+title: Calculating CAN traffic on a bus
+---
+
+
+
+In order to calculate the CAN load, we must declare a few variables and their importance within the CANbus protocol.
```
BlakeFreer, line 7: CAN, not CANbus

BlakeFreer (changes requested):

docs/docs/firmware/javascripts/katex.js
```diff
@@ -0,0 +1,12 @@
+document$.subscribe(({ body }) => { 
+
```
BlakeFreer, line 2: delete these extra lines

docs/docs/firmware/javascripts/katex.js
BlakeFreer, line 1: this file is not in the right location according to https://squidfunk.github.io/mkdocs-material/reference/math/#katex-docsjavascriptskatexjs

docs/mkdocs.yml
BlakeFreer, line 1: Still need to add some lines from https://squidfunk.github.io/mkdocs-material/reference/math/#katex-mkdocsyml

docs/docs/firmware/can-traffic.md
```diff
@@ -1,38 +1,46 @@
...
----
 
+# Calculating CAN traffic
 
+In order to calculate the CAN load, we must declare a few variables and their importance within the CAN protocol.
 
-In order to calculate the CAN load, we must declare a few variables and their importance within the CANbus protocol. 
-
-1. Baud: (kbaud)
+## Baud (kbaud)
```
BlakeFreer, line 6: Don't need the units here, an explanation of Baud is enough. Same with frequency

docs/docs/firmware/can-traffic.md
```diff
@@ -1,38 +1,46 @@
...
 
 The baud rate is the rate at which data is transfered in a communication channel (like CAN). For example, 9600 Baud rate is 9600 bits per second
 
-2. Frequency (Hz)
+## Frequency (Hz)
 
-Frequency in the CANbus refers to how many CAN frames (or messages) are transmitted per second. So the greater the frequency, the more bits transferred, increasing the bus load. 
+Frequency is the number of messages transmitted per second for each message type. Different messages can have different frequencies.
```
BlakeFreer, line 12: Can you reword this whole line like "The frequency of a message, in `Hz` is how many times it is transmitted per second."

docs/docs/firmware/can-traffic.md
```diff
@@ -1,38 +1,46 @@
...
 
-3. Total Data Per Cycle
+This variation affects the total bits transferred and, in turn, the bus load. To calculate bits per second, consider each message’s frequency and whether their lengths differ.
 
-Total data per cycle includes all the bits transmitted in one CAN message. This includes not only the payload but overhead bits like start of frame, CRC, acknowledgement, control field etc. 
+## Total Data Per Cycle
```
BlakeFreer, line 16: This should be called "Message length".

I would like you to calculate the message length as a function of the data length. You should specify the exact number of bits per message that come from the ID, Start of Frame, Data (dependent on data length), CRC, acknowledge, bit stuffing, end of frame, ex.

Use this site for the details: https://www.csselectronics.com/pages/can-bus-simple-intro-tutorial#can-bus-frames. You should recreate this diagram (or just download and cite it).

It includes all bit lengths BESIDES bit stuffing, which will be dependent on the data length. I'd like you to research what bit stuffing is and the maximum number of bits that it may add to a message.

Your final sentence should be "A message with data length of `x` bytes requires up to `a+bx` bits" where you have calculated `a` and `b`

docs/docs/firmware/can-traffic.md
```diff
@@ -1,38 +1,46 @@
...
+## The bus load is determined by
 
-## The bus load is determined by: 
+$$
+\text{Total Bits Per Second} = \sum_{i=1}^n \left(\text{Frequency}_i \times \text{Message Length}_i\right)
+$$
 
-![Bus Load Formula](https://latex.codecogs.com/png.latex?\text{Bus%20Load%20(\%)}%20=%20\frac{\text{Number%20of%20Bits%20per%20Frame}%20\times%20\text{Number%20of%20Frames%20per%20Second}}{\text{CAN%20Speed}}%20\times%20100)
+$$
+\text{Bus Load (%)} = \frac{\text{Total Bits per Second}}{\text{Baud Rate}} \times 100
```
BlakeFreer, line 27: add a percent sign with the `100`

docs/docs/firmware/can-traffic.md
```diff
@@ -1,38 +1,46 @@
...
+$$
 
-![Bus Load Formula](https://latex.codecogs.com/png.latex?\text{Bus%20Load%20(\%)}%20=%20\frac{\text{Total%20Bits%20per%20Second}}{\text{CAN%20Speed}}%20\times%20100)
+## Sample calculation
 
-## Sample calculation: 
+**Given**:
 
-Total data/cycle: 1524 (bits per cycle)
+Battery Status: Frequency = 100Hz, Length = 400 bits
```
BlakeFreer, line 34: after addressing the comment above, you should replace the `length` spec with `data length = x bytes` (you can choose x from 0-8) then calculate the number of bits from that. same for the next message

It would be cool to see this all in a table, like

| Message | Data Length (bytes) | Frequency (Hz) |
|:--------:|:-------:|:---:|
| BatteryStatus | 8 | 100 |
| MotorControl | 5 | 50 |

https://www.codecademy.com/resources/docs/markdown/tables

BlakeFreer (changes requested):

docs/docs/firmware/can-traffic.md
```diff
@@ -0,0 +1,80 @@
+
+# Calculating CAN traffic
```
BlakeFreer: capitalize "Traffic"

docs/docs/firmware/can-traffic.md
```diff
@@ -0,0 +1,80 @@
+
```
BlakeFreer: remove this blank line

docs/docs/firmware/can-traffic.md
```diff
@@ -0,0 +1,80 @@
...
+# Calculating CAN traffic
+
+In order to calculate the CAN load, we must declare a few variables and their importance within the CAN protocol.
```
BlakeFreer: replace "declare..." with "define a few variables related to the CAN protocol."

docs/docs/firmware/can-traffic/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+The frequency of a message, expressed in (Hz), is how often a message is transmitted per second. Different messages can have different frequencies.
+
+This variation affects the total bits transmitted and, in turn, the bus load. To calculate bits per second, consider each message’s frequency and whether their lengths differ.
+
+## Message Length
+
+The message length in a CAN frame can be calculated as a function of the data length. This calculation includes several fixed fields and a variable data field, which changes based on the message's data content.
+
+<figure markdown="span">
```
BlakeFreer, line 20: There is proper syntax for adding captions to an image but I didn't expect you to know it. I added it here for simplicity, just look it over. You can copy it from here if you ever need it.

docs/docs/firmware/can-traffic/index.md
```diff
@@ -0,0 +1,81 @@
...
+\text{Total Bits Per Second} = (8 \times 8 \times 100) + (5 \times 8 \times 50) = 6400 + 2000 = 8400 \, \text{bits per second}
+$$
+
+## Calculating Bus Load
+
+$$
+\text{Bus Load}\%= \frac{\text{Total Bits per Second}}{\text{Baud Rate}} \times 100\%
+$$
+
+$$
```
BlakeFreer, line 79: Not clear that this last equation is part of the previous example. It just looks like a continuation of the previous one. Just add a quick phrase before saying "The bus load in the previous example can be calculated as: "

docs/docs/firmware/can-traffic/index.md
BlakeFreer: This section is a little confusing. There is a lot of bouncing around between sentences and the equations aren't consistent. For example, you say `b=8` and `length = a+bx` but then also say `M = 44 + 10x`. Is `b` 8 or 10?

I think it would be better to restructure this as a table:

--- 

The CAN frame length is made up of some fixed elements and a variable data length. Let $x$ be the number of bytes in the data section.

| Frame Element | Length (bits) |
|:---|:---|
|Start of Frame | 1 |
| Identifier | 11 |
| Remote Transmission Request | 1 |
| ... | ... |
| Data | 8 bits per byte + up to 20% stuffing = 9.6x |
| and so on |

The maximum total frame length is $1 + 11 + 1 + 9.6x = 44 + 9.6x$ (fill in the other elements of the equation)

BlakeFreer: This will replace everything up to the Sample Calc header

docs/docs/firmware/can-traffic/index.md
```diff
@@ -0,0 +1,81 @@
...
+**Given**:
+
+Baud Rate: 500 kbaud (500,000 bits transferred per second)
+
+|Message Length| Data Length| Frequency (Hz)|
+|--------------|------------|---------------|
+|Battery Status|     8      |     100       |
+|Motor Control|      5      |     50        |
+
+Let \(M_1\) represent data length in bytes:
```
BlakeFreer, line 59: use `M_i` not `M_1` since this is generic for all messages

docs/docs/firmware/can-traffic/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+$$
+\text{Total Bits Per Second} = \sum_{i=1}^n \left(\text{Frequency}_i \times \text{Message Length}_i\right)
+$$
+
+$$
+\text{Total Bits Per Second} = (M_1 \times 8 \, \text{bits/byte} \times F_1) + (M_2 \times 8 \, \text{bits/byte} \times F_2)
```
BlakeFreer, line 66: you didn't use your equation here! the whole point of CAN traffic calcs is that there are more bit than just data bits going around

BlakeFreer: You could add # bits column to the table above. The first row would say `44 + 9.6 * 8 = 121`

docs/docs/firmware/can-traffic/index.md
ManushPatell: Would you like me to implement the syntax for adding captions. Would this be in regards to 
`Each CAN frame consists of several fields with fixed bit lengths, plus a variable-length data field. Let \(x\) be the number of bytes. Here’s the breakdown:` ?

docs/docs/firmware/can-traffic/index.md
```diff
@@ -22,60 +21,52 @@ The message length in a CAN frame can be calculated as a function of the data le
...
-
-**Calculation for fixed portion(\(a\))** : (\( 1 + 11 + 1 + 6 + 16 + 2 + 7 = 44\)) bits
+Each CAN frame consists of several fields with fixed bit lengths, plus a variable-length data field. Let \(x\) be the number of bytes. Here’s the breakdown:
 
-**Variable Portion (b)**: The data field adds 8 bits for each byte of data. So, \( b = 8 \).
-
-**Bit Stuffing**: To ensure synchronization, CAN adds around 20% bit stuffing to the frame length. Including bit stuffing, the approximate formulas are:
-
-- **For a Standard Frame**: \( M(x) \approx 44 + 10x \) bits
+|Frame Element| Length(bits)|
```
BlakeFreer, line 26: add space before (bits)

docs/docs/firmware/can-traffic/index.md
```diff
@@ -22,60 +21,52 @@ The message length in a CAN frame can be calculated as a function of the data le
...
+|--------------|------------|
+|Start of Frame (SOF) |     8   |
+|Identifier (ID)|      11     |
+|Remote Transmission Request (RTR)| 1|
+| Control (DLC)| 6|
+|Cyclic Redundancy Check (CRC) | 16 |
+|Acknowledgement (ACK)| 2 |
+|End of Frame (EOF)| 7|
+|Data Length| 8 bits per byte with 20% bit stuffing = 9.6x|
```
BlakeFreer, line 35: The frame element is called "Data" not "Data Length"

Also move the row so the table is in order.

docs/docs/firmware/can-traffic/index.md
```diff
@@ -22,60 +21,52 @@ The message length in a CAN frame can be calculated as a function of the data le
...
+Each CAN frame consists of several fields with fixed bit lengths, plus a variable-length data field. Let \(x\) be the number of bytes. Here’s the breakdown:
 
-**Variable Portion (b)**: The data field adds 8 bits for each byte of data. So, \( b = 8 \).
-
-**Bit Stuffing**: To ensure synchronization, CAN adds around 20% bit stuffing to the frame length. Including bit stuffing, the approximate formulas are:
-
-- **For a Standard Frame**: \( M(x) \approx 44 + 10x \) bits
+|Frame Element| Length(bits)|
+|--------------|------------|
+|Start of Frame (SOF) |     8   |
```
BlakeFreer, line 28: should this be 8?

docs/docs/firmware/can-traffic/index.md
```diff
@@ -22,60 +21,52 @@ The message length in a CAN frame can be calculated as a function of the data le
...
+| Control (DLC)| 6|
+|Cyclic Redundancy Check (CRC) | 16 |
+|Acknowledgement (ACK)| 2 |
+|End of Frame (EOF)| 7|
+|Data Length| 8 bits per byte with 20% bit stuffing = 9.6x|
 
 ### Final Formula
 
-A message with data length of \( x \) bytes requires up to \( a + bx \) bits.
+The maximum total frame length is: (\( 1 + 11 + 1 + 6 + 16 + 2 + 7 + 9.6x = 44 + 9.6x\)) bits
```
BlakeFreer: Don't need to put this expression in brackets. Since it's pretty long, put it on its own line by using `\[ \]` instead of `\( \)`. Make sure the order of terms matches the table

Would be good to remind the reader what `x` means here

docs/docs/firmware/can-traffic/index.md
```diff
@@ -22,60 +21,52 @@ The message length in a CAN frame can be calculated as a function of the data le
...
-$$
-
-$$
-\text{Total Bits Per Second} = (8 \times 8 \times 100) + (5 \times 8 \times 50) = 6400 + 2000 = 8400 \, \text{bits per second}
+\text{Total Bits Per Second} = (121 \times 100) + (92 \times 50) = 12100 + 4600 = 16700 \, \text{bits per second}
 $$
 
 ## Calculating Bus Load
 
+### The bus load is the previous example can be calculated as:
```
BlakeFreer, line 65: This line should be regular text, not a header, and it should go between the abstract equations for bus load and the example

BlakeFreer (changes requested):

docs/docs/firmware/can-traffic/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+Baud Rate: 500 kbaud (500,000 bits transferred per second)
+
+|Message Type | Data Length| Frequency (Hz)| Message Length (Bits)|
+|--------------|------------|---------------|---------------------|
+|Battery Status|     8      |     100       | 44 + 9.6 x 8 = 121|
+|Motor Control|      5      |     50        | 44 + 9.6 x 5 = 92|
+
+\begin{align}
+\text{Total Bits Per Second} &= \sum_{i=1}^n \left(\text{Frequency}_i \times \text{Message Length}_i \right)\\
```
BlakeFreer, line 57: This formula, and the bus load%, should be given before the example, after the final eq for message length. Add a header for "Total CAN Traffic" and "Bus Load Percentage" and give the equations there.

BlakeFreer (changes requested):

BlakeFreer (dismissed):
Looks good, I added the abbr extension back in since it's not causing any problems

PR #245, opened 2024-10-26
create add function

BlakeFreer:
@BlakeFreer can you look at this, not sure about that add function


firmware/shared/controls/add.cc
```diff
@@ -0,0 +1,3 @@
+int add(int x, int y) {
```
BlakeFreer, line 1: use better variable names

PR #252, opened 2024-10-26
STM32Cube Java Path on MacOS Incorrectly

### LINKED ISSUE #246: STM32Cube Java Path on MacOS Incorrectly Set
The command `make PROJECT=Demo/Blink PLATFORM=stm32f676 build` fails due to an incorrect Java path in `racecar/firmware/cmake/generate_cubemx.mk`.

The Java path should be set per OS, not globally.

Acceptance Criteria:

- Update Java path to be OS-specific.
- Ensure the command runs successfully on MacOS for cli and stm32f767.

`CUBEMX_JAVA := $(dir $(subst $(space),\$(space),$(CUBEMX_PATH)))jre/Contents/Home/bin/java`

CameronBeneteau:
Closes #246 

@BlakeFreer to verify build commands work on Windows.

`make PROJECT=Demo/Blink PLATFORM=cli build`
`make PROJECT=Demo/Blink PLATFORM=stm32f767 build`

Notes:
- Java path for MacOS assumes STM software was installed to the default location
- Possible issues if STM installation paths are different across Linux and Intel/Silicon Macs (will update in future if required)

### DIFF
--- a/firmware/.gitignore
+++ b/firmware/.gitignore
@@ -1,4 +1,7 @@
 **/build/
 /.clangd
 /.cache
+/.env
+/.venv
+/venv
 *.log
--- a/firmware/.gitignore
+++ b/firmware/.gitignore
@@ -1,7 +1,4 @@
 **/build/
 /.clangd
 /.cache
-/.env
-/.venv
-/venv
 *.log
--- a/firmware/cmake/generate_cubemx.mk
+++ b/firmware/cmake/generate_cubemx.mk
@@ -1,9 +1,20 @@
+# Find the STM32CubeMX executable
+ifeq ($(OS),Windows_NT)
+# Convert windows backslash to regular slash
+	CUBEMX_PATH := $(subst \,/,$(shell where STM32CubeMX))
+else
+# Linux / MacOS
+    CUBEMX_PATH := $(shell which STM32CubeMX)
+endif
+
+# Find the JAVA which is installed with CubeMX. Spaces in path are escaped.
 space := $(subst ,, )
-CUBEMX_PATH := $(shell which STM32CubeMX)
 
 ifeq ($(shell uname), Darwin)
+# MacOS
 	CUBEMX_JAVA := $(dir $(subst $(space),\$(space),$(CUBEMX_PATH)))jre/Contents/Home/bin/java
 else
+# Windows/Linux
 	CUBEMX_JAVA := $(dir $(subst $(space),\$(space),$(CUBEMX_PATH)))jre/bin/java
 endif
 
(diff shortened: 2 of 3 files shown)


firmware/cmake/generate_cubemx.mk
```diff
@@ -1,13 +1,14 @@
+space := $(subst ,, )
+
 ifeq ($(OS),Windows_NT)
 # Convert windows backslash to regular slash
 	CUBEMX_PATH := $(subst \,/,$(shell where STM32CubeMX))
+	CUBEMX_JAVA := $(dir $(subst $(space),\$(space),$(CUBEMX_PATH)))jre/bin/java
 else # Linux / MacOS
```
BlakeFreer, line 7: Linux is installed in the same path as windows, see https://github.com/macformula/racecar/pull/196

Your `if` should check for MacOS and have windows/linux be the else

BlakeFreer (changes requested):

BlakeFreer:
Also, since we specify git bash for windows users, we do not need to separate the `CUBEMX_PATH` expression. All platforms can use 
```makefile
CUBEMX_PATH := $(shell which STM32CubeMX)
```

firmware/cmake/generate_cubemx.mk
CameronBeneteau: Note: `$(OS)` on MacOS returns an empty string so need to check with `$(shell uname)`.

firmware/.gitignore
BlakeFreer, line 1: these new ignores should be added to `racecar/.gitignore` to ignore environments in all subfolders, not just in `firmware/`

firmware/cmake/generate_cubemx.mk
```diff
@@ -1,13 +1,12 @@
-ifeq ($(OS),Windows_NT)
-# Convert windows backslash to regular slash
-	CUBEMX_PATH := $(subst \,/,$(shell where STM32CubeMX))
-else # Linux / MacOS
-    CUBEMX_PATH := $(shell which STM32CubeMX)
+space := $(subst ,, )
+CUBEMX_PATH := $(shell which STM32CubeMX)
```
BlakeFreer, line 2: sorry, I made a mistake and didn't fully verify that `which` works on windows git bash. The function works and returns `.../STM32CubeMX` but `where` returns `.../STM32CubeMX.exe`. The Java call fails without the .exe extension.

Can you return this section as it was before? Don't combine it with the CUBEMX_JAVA if statement, I think it will be more clear with them handled separately

BlakeFreer (changes requested):

firmware/.gitignore
CameronBeneteau: Refactored. It would be nice to make the Python env naming consistent. Currently, documentation shows `.env` for firmware and `.venv` for docs. Not a big deal but could be a nice QoL change and first ticket for someone.

BlakeFreer (approved):

PR #256, opened 2024-10-27
driver interface error handling controls implementation

shene7:
temporary implementation for error handling, unsure if datatypes will always be double as simulink has it as auto


controls_rewrite_folder/driver_interface_error_handling.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    if (fractionIn >= 0 && fractionIn <= 1){
+        return fractionIn;
+    }
+    else{
+        return 0;
+    }
+}
+
+bool b_DriverInterfaceError(double DI_V_AccelPedalPos1, double DI_V_AccelPedalPos2, double DI_FrontBrakePressure, double DI_RearBrakePressure, double DI_V_RawSteeringAngle){
```
BlakeFreer, line 17: Great observation about the `double` datatype. We can "template" our functions to work on any datatype. Can you update both functions to use templates? https://en.cppreference.com/w/cpp/language/templates

controls_rewrite_folder/driver_interface_error_handling.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+
+    double PedalPos1 = potentiometerProcessing(DI_V_AccelPedalPos1);
+    double PedalPos2 = potentiometerProcessing(DI_V_AccelPedalPos2);
+    double FrontBrakePressure = potentiometerProcessing(DI_FrontBrakePressure);
+    double RearBrakePressure = potentiometerProcessing(DI_RearBrakePressure);
+    double RawSteeringAngle = potentiometerProcessing(DI_V_RawSteeringAngle);
+
+    return ((PedalPos1 == 0) || (PedalPos2 == 0) || (FrontBrakePressure == 0) || (RearBrakePressure == 0) || (RawSteeringAngle == 0) || (std::abs(DI_V_AccelPedalPos1 - DI_V_AccelPedalPos2) > 0.1));
```
BlakeFreer, line 26: This function will return an error if all doubles are 0, even if the sensors truly are 0.

This is where we will start to diverge from the simulink model in order to get our desired behaviour:

We should replace  `potentiometerProcessing` with a function `bool isInRange(value)` that returns true if it is between 0-1 and false otherwise. Then you can simplify the interface error function.

BlakeFreer (changes requested):

PR #257, opened 2024-10-27
Andrewi/236 jninja file loading

### LINKED ISSUE #236: cangen: Jinja file loading
Jinja2 can load templates directly from files, but cangen currently uses bare Python `open(template_path...`. We should use as much of Jinja's built in features, so fix this.

The Jinja2 docs likely have some examples for loading templates from a file

@AndrewI26 this issue should be tackled in a separate PR, so don't include it in #233

AndrewI26:
Changed _generate_from_jinja2_template to load templates directly from files.

closes #236

### DIFF
--- a/scripts/cangen/cangen/can_generator.py
+++ b/scripts/cangen/cangen/can_generator.py
@@ -13,7 +13,7 @@ from typing import Dict, List, Tuple
 
 import numpy as np
 from cantools.database import Database, Message, Signal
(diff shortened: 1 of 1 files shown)


BlakeFreer (approved):
Great work! I had no idea that Jinja had the package loader feature. It's perfect for our case and simplifies a lot of those CONSTANT paths.

PR #259, opened 2024-10-28
Changed template to match 'controls' template, added template to functions to account for unknown variables, cleaned up error handling function 

shene7:
Also added test cases, with tolerance to account for some particular cases that were not consistent. (for example, (abs(0.3 - 0.4) > 0.1 returned true, whereas abs(0.5-0.4) > 0.1 returned false). Compiled all test cases, with no failed cases. Let me know if I should add more / test more edge cases.


controls_rewrite_folder/driver_interface_error_handling.h
```diff
@@ -0,0 +1,20 @@
...
+template <typename T>
+bool isInRange(T fractionIn){
+
+    return (fractionIn >= 0 && fractionIn <= 1);
+}
+
+
+template <typename T>
+bool b_DriverInterfaceError(T DI_V_AccelPedalPos1, T DI_V_AccelPedalPos2, T DI_FrontBrakePressure, T DI_RearBrakePressure, T DI_V_RawSteeringAngle){
+    const T TOLERANCE = 0.00000000000001; // Define a small tolerance for floating-point comparisons
```
BlakeFreer, line 16: We do not need this tolerance since it does not matter if the pedals are exactly 0.1 apart.

controls_rewrite_folder/driver_interface_error_handling.h
```diff
@@ -0,0 +1,20 @@
...
+    return (!isInRange(DI_V_AccelPedalPos1) || !isInRange(DI_V_AccelPedalPos2) || !isInRange(DI_FrontBrakePressure) || !isInRange(DI_RearBrakePressure) || !isInRange(DI_V_RawSteeringAngle) || (std::abs(DI_V_AccelPedalPos1 - DI_V_AccelPedalPos2) > TOLERANCE+ 0.1));
```
BlakeFreer, line 17: just check if abs(...) > 0.1, forget the tolerance

controls_rewrite_folder/driver_interface_error_handling_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+    // Test case 5: RearBrakePressure out of range
+    ASSERT_EQ(b_DriverInterfaceError(0.8, 0.7, 0.9, -0.1, 0.5), true);
+
+    // Test case 6: RawSteeringAngle out of range
+    ASSERT_EQ(b_DriverInterfaceError(0.8, 0.7, 0.9, 0.8, -0.5), true);
+
+    // Test case 7: Difference between PedalPos1 and PedalPos2 exceeds threshold
+    ASSERT_EQ(b_DriverInterfaceError(0.8, 0.95, 0.9, 0.8, 0.5), true);
+
+    // Test case 8: Pedal positions difference exactly on threshold
```
BlakeFreer, line 30: remove this test, behaviour at the boundary does not matter

controls_rewrite_folder/driver_interface_error_handling_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+    ASSERT_EQ(b_DriverInterfaceError(0.8, 0.9, 0.9, 0.8, 0.5), false);
+
+    // Test case 9: Pedal position and parameters should NOT result in true
```
BlakeFreer, line 33: if you're going to assert NEQ true, just use EQ false, its easier to read. Same for below

BlakeFreer (changes requested):
I added two new asserts to dev/controls which are `ASSERT_TRUE` and `ASSERT_FALSE`. I should have added them before since they are really useful for testing boolean expressions like your function. Please update to use them.

There seems to be a misunderstanding about how to organize the project. Your files should be placed in the `firmware/shared/controls` directory, not `controls_rewrite_folder`, and you need to `rebase` your onto the `dev` branch to see those files. To do this, follow these steps.

```bash
git checkout dev/controls
git pull
git checkout user/evshe_feature
git rebase dev/controls
```

This will apply your changes on top of the `dev/controls` branch.

Then, move your files into the correct folder. You must execute these commands from `racecar/`

```bash
git mv controls_rewrite_folder/* firmware/shared/controls
```

It would be good to commit your changes now. when you go to push, you will need to `git push --force` since we changed the structure of your git history with `rebase`

You should delete your `_testing.h` file since your files will now be alongside the files that I made earlier, so you can use that `testing.h` file

controls_rewrite_folder/driver_interface_error_handling_test.cpp
BlakeFreer: also the other tests which check for exactly 0.1

BlakeFreer:
Can you install `pre-commit` using these instructions? Then when you commit your code, it will auto-format it to make it look consistent (like even whitespace and such).

To get the formatter to run for this first time, edit both of your files (simply add an extra blank line somewhere) then `git add` and `git commit` them. Your commit _should_ be rejected by pre-commit with a message like

```bash
$ git commit -m "fmt"
don't commit to branch...................................................Passed
Run git-clang-format.....................................................Failed
- hook id: clang-format
- exit code: 1
- files were modified by this hook

changed files:
    firmware/shared/controls/driver_interface_error_handling.h    
    firmware/shared/controls/driver_interface_error_handling_test.cc
```

If you look at the files then they will look different and (hopefully) nicer. Formatting helps us keep all of our code looking nice.

After that failed commit, re-add the files (with their formatting changes) and commit again. It should succeed on the second or third round

BlakeFreer (changes requested):

firmware/shared/controls/driver_interface_error_handling.h
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
+#include <cmath>
+#pragma once
```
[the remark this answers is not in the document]
shene7: oops thought I did that guess I didn't save it

BlakeFreer (changes requested):
Still need to delete the old `.cpp` file.

The `_test` file does not compile. Make sure you test it before making a PR

```bash
make driver_interface_error_handling_test
```

BlakeFreer (changes requested):
Still need to delete the old `.cpp` file.

The `_test` file does not compile. Make sure you test it before making a PR

```bash
make driver_interface_error_handling_test
```

BlakeFreer (approved):
looks good!

PR #272, opened 2024-10-31
controls: Torque Request

ojesene1:


firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <cmath>
+#include <iostream>
```
BlakeFreer: You can remove the iostream line since it's not used by this header file

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <cmath>
+#include <iostream>
+
+namespace ctrl {  
+
+// Templating means this function will work on any types that support addition.
+template <typename T, typename B>
```
BlakeFreer: We expect READY_TO_DRIVE and DI_Error to be booleans, so they should not be templated. The accelpedal pos should be templated as you have it.

firmware/shared/controls/DI_driverTorqueRequest_test.cpp
BlakeFreer, line 1: this file should use the `.cc` suffix. To change this, run

```bash
git mv DI_driverTorqueRequest_test.cpp DI_driverTorqueRequest_test.cc
```

Then the makefile should work by running 

```bash
make DI_driverTorqueRequest_test
```

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -0,0 +1,26 @@
...
+// Templating means this function will work on any types that support addition.
+template <typename T, typename B>
+T DI_driverTorqueRequest(T READY_TO_DRIVE, T DI_Error, B AccelPedPos) {
+    T u1 = 0.0;
+    if((!READY_TO_DRIVE) == 1 || DI_Error == 1){ 
+        return u1;
+    }
+    else{ 
+        //AccelPedPos (n-d Lookup Table)
+        if(AccelPedPos > 1){ return 1.0;}
```
BlakeFreer: this `if / elseif / else` block always returns either 0 or 1, but this is not the behaviour of the original function.

The simulink function just returns the accel_pedal_pos (see the simulink diagram)

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -0,0 +1,26 @@
...
+#include <cmath>
+#include <iostream>
+
+namespace ctrl {  
+
+// Templating means this function will work on any types that support addition.
+template <typename T, typename B>
+T DI_driverTorqueRequest(T READY_TO_DRIVE, T DI_Error, B AccelPedPos) {
+    T u1 = 0.0;
+    if((!READY_TO_DRIVE) == 1 || DI_Error == 1){
```
BlakeFreer: since these two variables will be made to booleans, you should replace `== 1` with `== true`
However, `X ==true` is equivalent to just `X`, so this line can be simplified even more

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <cmath>
+#include <iostream>
+
+namespace ctrl {  
+
+// Templating means this function will work on any types that support addition.
+template <typename T, typename B>
+T DI_driverTorqueRequest(T READY_TO_DRIVE, T DI_Error, B AccelPedPos) {
+    T u1 = 0.0;
```
BlakeFreer: since this variable is always 0, it is more clear to just return 0 and get rid of this value. This makes the code simpler to read since the reader doesn't need to mentally "remember" what the `u1` variable means.

This is an example of where we will take a C++ approach instead of using the exact simulink syntax.

BlakeFreer (changes requested):

BlakeFreer:
I didn't mentioned this earlier, but we have a script which automatically formats your code when you `git commit` it. This keeps everything looking clean and consistent. You can install it with these pre-commit instructions https://macformula.github.io/racecar/firmware/dev-setup/#pre-commit-setup

When you make a commit, the script will try to format your code. If it makes any changes. your commit will fail, so then you just need to `git add` the changed files again and re-run `git commit`

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -1,25 +1,14 @@
 #pragma once
 #include <cmath>
```
BlakeFreer, line 2: don't need `cmath` anymore since you aren't using `round()`

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -1,25 +1,14 @@
...
-
 namespace ctrl {
 
 // Templating means this function will work on any types that support addition.
-template <typename T, typename B>
-T DI_driverTorqueRequest(T READY_TO_DRIVE, T DI_Error, B AccelPedPos) {
-    T u1 = 0.0;
-    if ((!READY_TO_DRIVE) == 1 || DI_Error == 1) {
-        return u1;
+template <typename B>
```
BlakeFreer, line 6: your templates look good, only thing is to rename `B` to the more standard template variable `T`

firmware/shared/controls/DI_driverTorqueRequest.h
```diff
@@ -1,25 +1,14 @@
...
+B DI_driverTorqueRequest(bool READY_TO_DRIVE, bool DI_Error, B AccelPedPos) {
```
BlakeFreer, line 7: All parameters should be spelled with `lower_snake_case` to follow our naming conventions. See the [C++ Style Guide -> Variable Names](https://google.github.io/styleguide/cppguide.html#Variable_Names)

Is there a more descriptive name for the error? Where does it come from in the Simulink model / what error does it represent?

firmware/shared/controls/DI_driverTorqueRequest_test.cc
```diff
@@ -10,19 +10,16 @@ int main() {
...
+    // should return AccelPedPos = 0.0006f
+    ASSERT_CLOSE(ctrl::DI_driverTorqueRequest(true, false, 0.0006f), 0.0006f);
 
-    // should return AccelPedPos = 0
-    ASSERT_NCLOSE(ctrl::DI_driverTorqueRequest(true, false, 0.4), 2.0);
-
-    // should return AccelPedPos = 0 (negative position case)
-    ASSERT_EQ(ctrl::DI_driverTorqueRequest(true, false, -1), 0);
+    // should return AccelPedPos = 0.4
+    ASSERT_NCLOSE(ctrl::DI_driverTorqueRequest(true, false, 0.4), 0.5);
```
BlakeFreer, line 23: Can you add a test with boolean parameters `true, true`? It's the only bool combination not tested

BlakeFreer (changes requested):

firmware/shared/controls/DI_driverTorqueRequest_test.cc
BlakeFreer, line 1: remove this old file

BlakeFreer, line 1: remove this old file

firmware/shared/controls/driver_interface_torque_req_test.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "driver_interface_torque_req.h"  // include the header with your function implementation
+
+#include <iostream>
+
+#include "testing.h"
+
+int main() {
+    // Write your test cases.
+
+    // should return 0
```
BlakeFreer, line 12: For these comments, I can see what the test "should" return from the ASSERT statement. A more useful comment is explaining why it should be 0 (in this case, because `ready_to_drive` is false)

firmware/shared/controls/driver_interface_torque_req_test.cc
```diff
@@ -0,0 +1,30 @@
...
+
+    // should return accel_ped_pos = 2.5
+    ASSERT_CLOSE(ctrl::driver_interface_torque_req(true, false, 2.5), 2.5);
+
+    // should return accel_ped_pos = 0.0006f
+    ASSERT_CLOSE(ctrl::driver_interface_torque_req(true, false, 0.0006f),
+                 0.0006f);
+
+    // should return accel_ped_pos = 0.4
+    ASSERT_NCLOSE(ctrl::driver_interface_torque_req(false, true, 0.4), 0.5);
```
BlakeFreer, line 23: This test is "stronger" if we test for equality with `0.4` rather than inequality with some other value. For example, if your function returned 0.6 instead of 0.4, it would still pass.

You do not need to use an `_NCLOSE` test here if _CLOSE is better suited

BlakeFreer (changes requested):

ojesene1:
I pushed the updated test file and deleted the previous files.

BlakeFreer (approved):

Issue #239, opened 2024-10-23
cangen: update config


BlakeFreer:
Update cangen/config.py and all project config.yamls to accommodate these changes:

1. Rename `busName` to `name` in yaml and `bus.bus_name` to `bus.name` in py. We are clearly in the context of a bus so no reason to repeat it
2. Specify `ourNode` per bus. Currently, dbc files are coupled because they must use the same node name (ex pt.dbc and veh.dbc must both use `FC`.) This coupling is unnecessary. Also remove "our" from the yaml and python variable names

BlakeFreer:
closed by #276

PR #276, opened 2024-11-01
changed config names and added node to bus #239

AndrewI26:


BlakeFreer (approved):
tested on front controller, gives the same output as `main`

PR #277, opened 2024-11-02
fix: :bug: Fix TMS halt and temperature calculations.

BlakeFreer:
Previously, TMS would appear to crash upon boot. This was caused by the `HAL_CAN_RxFifo0MsgPendingCallback` interrupt being triggered when the BMS sent CAN messages, even though the TMS does not care about them. TMS did not call `HAL_CAN_GetRxMessage()` in the interrupt to clear the interrupt flag, so the interrupt continued to fire, preventing the main code from running. This commit fixes the problem by handling the messages.

Also fixes the temperature calculation which is now explicitly calculated from the sensor datasheet table and the physical hardware circuit.

As far as I can tell, TMS "works" now. It boots, sends messages, and correctly calculates the temperature values.

Verified the temp calc by reading the CAN message which stated the battery temp was 19 degC which matched the Hatch thermostat.

### DIFF
--- a/firmware/projects/EV5/TMS/platforms/stm32f767/bindings.cc
+++ b/firmware/projects/EV5/TMS/platforms/stm32f767/bindings.cc
@@ -12,7 +12,6 @@
 #include "main.h"
 #include "mcal/stm32f767/periph/can.h"
 #include "shared/comms/can/can_msg.h"
-#include "stm32f7xx_hal_tim.h"
 #include "tim.h"
 
 // fw imports
--- a/firmware/projects/EV5/TMS/platforms/stm32f767/bindings.cc
+++ b/firmware/projects/EV5/TMS/platforms/stm32f767/bindings.cc
@@ -1,6 +1,8 @@
 /// @author Blake Freer
 /// @date 2023-12-25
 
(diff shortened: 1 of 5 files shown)


CameronBeneteau (dismissed):
👏👏

BlakeFreer:
I want to do one more test on monday to make sure nothing got messed up when I rebased my commits

firmware/projects/EV5/TMS/platforms/stm32f767/bindings.cc
```diff
@@ -10,6 +12,7 @@
 #include "main.h"
 #include "mcal/stm32f767/periph/can.h"
 #include "shared/comms/can/can_msg.h"
+#include "stm32f7xx_hal_tim.h"
```
BlakeFreer, line 15: Remove this

TylerStAmour (approved):
Lgtm

BlakeFreer:
Let's go! This was the last ECU of EV5 to be moved to `racecar/`! We're officially a gen2 firmware car. Thanks @CameronBeneteau for getting openocd working, we wouldn't have been able to debug without it.

PR #281, opened 2024-11-02
Andrewi/264 output folder prep

AndrewI26:


scripts/cangen/cangen/can_generator.py
```diff
@@ -247,13 +237,25 @@ def generate_code(bus: Bus, config: Config):
     logger.info("Code generation complete")
 
 
+def _prepare_output_directory(output_dir):
+    if os.path.exists(output_dir):
+        shutil.rmtree(output_dir)
+
+    os.makedirs(output_dir, exist_ok=True)
```
BlakeFreer, line 244: these commands can be rearranged

1. make the folder, dont raise error if it exists
2. clear the folder with shutil, already know the folder exists since we just created it
3. create the gitignore

you don't need to check if the gitignore exists because we just cleaned the folder

scripts/cangen/cangen/can_generator.py
```diff
@@ -247,13 +237,25 @@ def generate_code(bus: Bus, config: Config):
...
+
+
 def generate_can_from_dbc(project_folder_name: str):
     os.chdir(project_folder_name)
     config = Config.from_yaml("config.yaml")
 
-    # Deletes output path folder and files within, before creating new ones
-    if os.path.exists(config.output_dir):
-        shutil.rmtree(config.output_dir)
+    # Deletes previously generated files and creates a gitignore for the directory
```
BlakeFreer, line 257: this comment would be better as a docstring comment for the function https://peps.python.org/pep-0257/#one-line-docstrings since it relates more to the behaviour of the function rather than the context where it is called

BlakeFreer (changes requested):

scripts/cangen/cangen/can_generator.py
```diff
@@ -247,13 +237,22 @@ def generate_code(bus: Bus, config: Config):
     logger.info("Code generation complete")
 
 
+def _prepare_output_directory(output_dir):
+    """Deletes previously generated files and creates a gitignore for the directory"""
+    os.makedirs(output_dir, exist_ok=True)
+    shutil.rmtree(output_dir)
+    
+    os.makedirs(output_dir, exist_ok=True)
```
BlakeFreer, line 245: Is this line needed?

AndrewI26: rmtree removes the contents of the folder and the folder itself, but since the folder directory we give it is generated/can, it removes the /can directory and everything after it. When I then try to write the gitignore, it throws an error because the generated directory does not contain the /can path afterwords. If we wanted to delete the contents of the generated/can directory, without deleting the folder itself, we can loop over its contents then delete each path. This way is a little more verbose but definitely more readable.

scripts/cangen/cangen/can_generator.py
```diff
@@ -247,13 +237,22 @@ def generate_code(bus: Bus, config: Config):
     logger.info("Code generation complete")
 
 
+def _prepare_output_directory(output_dir):
+    """Deletes previously generated files and creates a gitignore for the directory"""
+    os.makedirs(output_dir, exist_ok=True)
+    for file in os.listdir(output_dir):
+        os.unlink(file)
```
BlakeFreer, line 244: `file` is relative to `output_dir` but `os.unlink` is relative to the current dir (ex. projects/EV5/TMS) so this removes the wrong file. Also unlink will not work if `generated/can` contains folders at some point

I think we should go back to something like what we had before:
```
if (output dir exists):
    rmtree(output_dir)

makedirs(output_dir)
create .gitignore
```

sorry for going backwards, i assumed there was a simple function for clearing a folder without deleting it but that doesn't seem to be the case.

BlakeFreer (changes requested):

BlakeFreer (approved):
looks good, tested

PR #287, opened 2024-11-03
Shene7/feature

shene7:
Wrote general features and created some test cases. Test cases succeed and compiles, let me know if there's anything I should change

### DIFF
--- a/firmware/shared/controls/tvFactor.h
+++ b/firmware/shared/controls/tvFactor.h
@@ -5,34 +5,25 @@
 
 namespace ctrl{
 
-bool EnableTorqueVectoring(bool torque_vectoring_switch) {
-    if (torque_vectoring_switch == true) {
-        return true;
-    } else {
-        return false;
-    }
-}
-
 template <typename T>
-T CreateTorqueVectoringFactor(T steering_angle, bool torque_vectoring_switch) {
-    if (EnableTorqueVectoring(torque_vectoring_switch) == true) {
+T CreateTorqueVectoringFactor(T steering_angle, bool enable_torque_vectoring) {
+    if (enable_torque_vectoring == false) {
+        return 0;
+    } else {
         T absolute_steering_angle = std::abs(steering_angle);
 
-        static const float steering_angle_lookup_table_data[][2] = {
+        static const float table_data[][2] = {
             {0.0, 1.0f},    {5.0, 0.934f},  {10.0, 0.87f},
             {15.0, 0.808f}, {20.0, 0.747f}, {25.0, 0.683f},
         };
 
-        static constexpr int steering_angle_lookup_table_length =
-            (sizeof(steering_angle_lookup_table_data)) /
-            (sizeof(steering_angle_lookup_table_data[0]));
+        static constexpr int table_length =
+            (sizeof(table_data)) / (sizeof(table_data[0]));
 
-        static shared::util::LookupTable<steering_angle_lookup_table_length>
-            table_lookup{steering_angle_lookup_table_data};
+        static shared::util::LookupTable<table_length> tv_lookup_table{
+            table_data};
 
-        return table_lookup.Evaluate(absolute_steering_angle);
-    } else {
-        return 0;
+        return tv_lookup_table.Evaluate(absolute_steering_angle);
     }
 }
 
--- a/firmware/shared/controls/tvFactor.h
+++ b/firmware/shared/controls/tvFactor.h
@@ -6,25 +6,20 @@
 namespace ctrl{
 
 template <typename T>
-T CreateTorqueVectoringFactor(T steering_angle, bool enable_torque_vectoring) {
-    if (enable_torque_vectoring == false) {
-        return 0;
-    } else {
-        T absolute_steering_angle = std::abs(steering_angle);
+T CreateTorqueVectoringFactor(T steering_angle) {
+    T absolute_steering_angle = std::abs(steering_angle);
 
-        static const float table_data[][2] = {
-            {0.0, 1.0f},    {5.0, 0.934f},  {10.0, 0.87f},
-            {15.0, 0.808f}, {20.0, 0.747f}, {25.0, 0.683f},
-        };
+    static const float table_data[][2] = {
+        {0.0, 1.0f},    {5.0, 0.934f},  {10.0, 0.87f},
+        {15.0, 0.808f}, {20.0, 0.747f}, {25.0, 0.683f},
+    };
 
-        static constexpr int table_length =
-            (sizeof(table_data)) / (sizeof(table_data[0]));
+    static constexpr int table_length =
+        (sizeof(table_data)) / (sizeof(table_data[0]));
 
-        static shared::util::LookupTable<table_length> tv_lookup_table{
-            table_data};
+    static shared::util::LookupTable<table_length> tv_lookup_table{table_data};
 
-        return tv_lookup_table.Evaluate(absolute_steering_angle);
-    }
+    return tv_lookup_table.Evaluate(absolute_steering_angle);
 }
 
 template <typename T>
--- a/firmware/shared/controls/tvFactor.h
+++ b/firmware/shared/controls/tvFactor.h
@@ -23,13 +23,13 @@ T open_loop_tv_lookUp(T DI_p_steeringAngle){
     // breakpoints = {0, 5, 10, 15, 20, 25}; 
     // tableData = {1, 0.934, 0.87, 0.808, 0.747, 0.683};
 
-    const T steeringAngle_lut_data[][2] = {
-    {0, 1},
-	{5, 0.934},
-	{10, 0.87},
-	{15, 0.808},
-	{20, 0.747},
-	{25, 0.683},
+    const float steeringAngle_lut_data[][2] = {
+    {0.0, 1.0f},
+	{5.0, 0.934f},
+	{10.0, 0.87f},
+	{15.0, 0.808f},
+	{20.0, 0.747f},
+	{25.0, 0.683f},
     };
 
     constexpr int steeringAngle_lut_length =
@@ -37,25 +37,28 @@ T open_loop_tv_lookUp(T DI_p_steeringAngle){
 
     shared::util::LookupTable<steeringAngle_lut_length> steeringAngle_adc_lut{steeringAngle_lut_data};
 
-    T result = steeringAngle_adc_lut.get(DI_p_steeringAngle);
+    T result = steeringAngle_adc_lut.Evaluate(Angle);
+    return result;
 }
 
+
+        
 template <typename T>
 std::tuple<T,T> tvFactoring(T p_steeringAngle, T tvFactor){    //don't pass by reference (struct or tuple)
     T p_tvFactorLeft;
     T p_tvFactorRight; 
 
     if (p_steeringAngle > 0){
-        p_tvFactorLeft = 1;
+        p_tvFactorLeft = T(1);
         p_tvFactorRight = tvFactor;
     }
     else if (p_steeringAngle < 0){
-        p_tvFactorRight = 1;
+        p_tvFactorRight = T(1);
         p_tvFactorLeft = tvFactor;
     }
     else{
-        p_tvFactorLeft = 1;
-        p_tvFactorRight = 1;
+        p_tvFactorLeft = T(1);
+        p_tvFactorRight = T(1);
     }
 
     return std::make_tuple(p_tvFactorLeft, p_tvFactorRight);
--- a/firmware/shared/controls/tvFactor.h
+++ b/firmware/shared/controls/tvFactor.h
@@ -2,6 +2,8 @@
 #include <cmath>
 #include <vector>
 #include <algorithm>
+#include <tuple>
+#include "../util/mappers/lookup_table.h"
 
 namespace ctrl{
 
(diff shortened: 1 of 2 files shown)


firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+#include <cmath>
```
BlakeFreer, line 2: remove unused imports

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+#include <vector>
+#include <algorithm>
+#include <tuple>
+#include "../util/mappers/lookup_table.h"
+
+namespace ctrl{
+
+template <typename T>
+T tvEnable(T DI_p_steeringAngle, T DI_b_tvEnable){
```
BlakeFreer, line 11: should tvEnable be a boolean?

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+
+    if (DI_b_tvEnable > 0) return DI_p_steeringAngle;
+    else return 0;
+}
+
+template <typename T>
+T open_loop_tv_lookUp(T DI_p_steeringAngle){
```
BlakeFreer, line 18: fix variable & function names

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+
+    T Angle = std::abs(DI_p_steeringAngle);
```
BlakeFreer, line 20: variable anme

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+
+    // Lookup Table Data
+    // breakpoints = {0, 5, 10, 15, 20, 25}; 
+    // tableData = {1, 0.934, 0.87, 0.808, 0.747, 0.683};
+
+    const float steeringAngle_lut_data[][2] = {
```
BlakeFreer, line 26: code is not formatted. install `pre-commit`

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+	{20.0, 0.747f},
+	{25.0, 0.683f},
+    };
+
+    constexpr int steeringAngle_lut_length =
+    (sizeof(steeringAngle_lut_data)) / (sizeof(steeringAngle_lut_data[0]));
+
+    shared::util::LookupTable<steeringAngle_lut_length> steeringAngle_adc_lut{steeringAngle_lut_data};
+
+    T result = steeringAngle_adc_lut.Evaluate(Angle);
```
BlakeFreer, line 40: this `T result = ...` and `return result` can be inlined, no need to create a variable

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+    return result;
+}
+
+template <typename T>
+std::tuple<T,T> tvFactoring(T p_steeringAngle, T tvFactor){    //don't pass by reference (struct or tuple)
+    T p_tvFactorLeft;
+    T p_tvFactorRight; 
+
+    if (p_steeringAngle > 0){
+        p_tvFactorLeft = T(1);
```
BlakeFreer, line 50: use `static_cast<T>( )` instead of  `T( )`

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+    else if (p_steeringAngle < 0){
+        p_tvFactorRight = T(1);
+        p_tvFactorLeft = tvFactor;
+    }
+    else{
+        p_tvFactorLeft = T(1);
+        p_tvFactorRight = T(1);
+    }
+
+    return std::make_tuple(p_tvFactorLeft, p_tvFactorRight);
```
BlakeFreer, line 62: use `std::tuple` instead of maketuple

firmware/shared/controls/tvFactor_test.cc
```diff
@@ -0,0 +1,42 @@
+#include <iostream> 
+
+#include "tvFactor.h"
+
+#include "testing.h"
+
+int main (){
+    // Use these for float / double types
+    // #define ASSERT_CLOSE(x, y) assert(std::abs((x) - (y)) < (decltype(x))TOLERANCE)
```
BlakeFreer, line 9: remove these comments

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+    (sizeof(steeringAngle_lut_data)) / (sizeof(steeringAngle_lut_data[0]));
+
+    shared::util::LookupTable<steeringAngle_lut_length> steeringAngle_adc_lut{steeringAngle_lut_data};
+
+    T result = steeringAngle_adc_lut.Evaluate(Angle);
+    return result;
+}
+
+template <typename T>
+std::tuple<T,T> tvFactoring(T p_steeringAngle, T tvFactor){    //don't pass by reference (struct or tuple)
```
BlakeFreer, line 45: var and func names

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,65 @@
...
+    T p_tvFactorLeft;
+    T p_tvFactorRight; 
+
+    if (p_steeringAngle > 0){
```
BlakeFreer, line 49: this logic doesn't look correct. What is the simulink model doing? Seems strange to have such a discontinuous torque response

BlakeFreer (changes requested):
Your branch name is not descriptive. It should be something like shene7/torque_vectoring. When I said to call your branch "user/feature" I meant that you should replace __both__ the user with your name and feature with your topic. don't need to change anything for this branch but make sure to do it in the future

firmware/shared/controls/tvFactor.h
shene7: That's a very good point. When I looked at the block originally, I wasn't sure since it seems the intention of if statement block was to specifically pass either 0 or the steeringAngle itself to the open_loop_tv block. I also thought: is it a waste to have this function in the first place? All this function does is see if the enable variable is over 0 or not, to which it sends the steering angle or 0 to the lookup table function. If it's going to be a boolean that just enables the use of another function, perhaps instead you would have just a boolean variable that updates on a certain condition elsewhere in the code.

firmware/shared/controls/tvFactor.h
shene7: hmm, the matlab code is as follows: am I missing something? 

```function [p_tvFactorLeft, p_tvFactorRight]= fcn(p_tvFactor, p_steeringAngle)
%Assuming right is positive
if p_steeringAngle > 0
    p_tvFactorLeft = 1;
    p_tvFactorRight = p_tvFactor;
elseif p_steeringAngle < 0
    p_tvFactorRight = 1;
    p_tvFactorLeft = p_tvFactor;
else
    p_tvFactorLeft = 1;
    p_tvFactorRight = 1;
end```

Comment (BlakeFreer) [reply firmware/shared/controls/tvFactor.h]: ok I see, i forgot that p_tv is pass into the function

File: firmware/shared/controls/tvFactor.h (hunk 2 above)
Comment (BlakeFreer) [reply firmware/shared/controls/tvFactor.h]: the simulink var name is `DI_b...` and the b is boolean, so let's make this a bool

File: firmware/shared/controls/tvFactor.h [hunk 12]
```diff
@@ -0,0 +1,54 @@
+#pragma once
+#include <vector>
```
BlakeFreer, line 2: unused

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,54 @@
...
+#include <algorithm>
+#include <tuple>
+#include "../util/mappers/lookup_table.h"
+
+namespace ctrl{
+
+template <typename T>
+T EnableTorqueVectoring(T steering_angle, T torque_vectoring_switch) {
+    if (torque_vectoring_switch > 0)
+        return steering_angle;
```
BlakeFreer, line 12: add `{ }` braces around these statements

firmware/shared/controls/tvFactor.h
```diff
@@ -0,0 +1,54 @@
...
+    else return 0;
+}
+
+template <typename T>
+T CreateTorqueVectoringFactor(T steering_angle) {
+    T absolute_steering_angle = std::abs(steering_angle);
+
+    const float steering_angle_lookup_table_data[][2] = {
```
BlakeFreer, line 20: since these 3 variables related to the lookup table are constant for each function call, it is more efficient to make them static variables. I''d like you to learn why and make this change

firmware/shared/controls/tvFactor_test.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "tvFactor.h"
+
+#include "testing.h"
+
+int main() {
+    using namespace ctrl;
+
+    ASSERT_CLOSE(
+        EnableTorqueVectoring(26.3, 0.2),
+        26.3);  // Should return 30.0 because DI_b_EnableTorqueVectoring > 0
```
BlakeFreer, line 12: comment doesn't match

firmware/shared/controls/tvFactor_test.cc
```diff
@@ -0,0 +1,44 @@
...
+        EnableTorqueVectoring(12.0, -6.2),
+        0.0);  // Should return 0.0 because DI_b_EnableTorqueVectoring <= 0
+
+    ASSERT_CLOSE(CreateTorqueVectoringFactor(5.0), 0.934);
+    ASSERT_CLOSE(CreateTorqueVectoringFactor(10.0), 0.87);
+    ASSERT_CLOSE(CreateTorqueVectoringFactor(13.2), 0.83032);
+    ASSERT_CLOSE(CreateTorqueVectoringFactor(23.3), 0.70476);
+    ASSERT_CLOSE(CreateTorqueVectoringFactor(27.3), 0.683);
+
+    auto [leftFactor1, rightFactor1] = AdjustTorqueVectoring(15.2, 3.6);
```
BlakeFreer, line 29: use scoping notation since you're creating variables. I know it's not a problem for this set of tests, but it's good practice to "delete" the variables after they go out of scope. 

To do this, wrap lines 29-31 in `{ }` like this

```c++
{
    auto [leftFactor1, rightFactor1] = AdjustTorqueVectoring(15.2, 3.6);
    ASSERT_CLOSE(leftFactor1, 1.0); 
    ASSERT_CLOSE(rightFactor1, 3.6);
}
```

Then the left/right factor variables are not accessible after the block of code, so you don't even need to number them. Also fix the variable naming as per the guidelines

BlakeFreer (changes requested):

firmware/shared/controls/tvFactor.h
```diff
@@ -1,35 +1,39 @@
...
 #include "../util/mappers/lookup_table.h"
 
 namespace ctrl{
 
-template <typename T>
-T EnableTorqueVectoring(T steering_angle, T torque_vectoring_switch) {
-    if (torque_vectoring_switch > 0)
-        return steering_angle;
-    else return 0;
+bool EnableTorqueVectoring(bool torque_vectoring_switch) {
```
BlakeFreer, line 8: why have this function? it just returns its input

firmware/shared/controls/tvFactor.h
```diff
@@ -1,35 +1,39 @@
...
+        return true;
+    } else {
+        return false;
+    }
 }
 
 template <typename T>
-T CreateTorqueVectoringFactor(T steering_angle) {
-    T absolute_steering_angle = std::abs(steering_angle);
+T CreateTorqueVectoringFactor(T steering_angle, bool torque_vectoring_switch) {
```
BlakeFreer, line 17: a better name for the bool is `enable_torque_vectoring` since "switch" does not indicate what true/false means

firmware/shared/controls/tvFactor.h
```diff
@@ -1,35 +1,39 @@
...
+    if (EnableTorqueVectoring(torque_vectoring_switch) == true) {
```
BlakeFreer, line 18: this whole `if` block should be reversed to make it clearer what happens on `false`:

```c++
if(not enabled) {
    return 0
}

static table = [][]...

return table.evaluate()

File: firmware/shared/controls/tvFactor.h [hunk 20]
```diff
@@ -1,35 +1,39 @@
...
+        T absolute_steering_angle = std::abs(steering_angle);
 
-    const float steering_angle_lookup_table_data[][2] = {
-        {0.0, 1.0f},    {5.0, 0.934f},  {10.0, 0.87f},
-        {15.0, 0.808f}, {20.0, 0.747f}, {25.0, 0.683f},
-    };
+        static const float steering_angle_lookup_table_data[][2] = {
```
BlakeFreer, line 21: it is very clear from the context what the variables refer to, so they don't need to be so long. simpler names are

`table_data`, `table_length` and `tv_lookup_table`

BlakeFreer (changes requested):

firmware/shared/controls/tvFactor.h
shene7: That's exactly what I thought to be honest, but I wasn't sure if I should omit it since we talked previously about making the function a bool

BlakeFreer (approved):

PR #292, opened 2024-11-04
Controls: TC Scale Factor

ojesene1:
Created the function and test file for p_tcScaleFactor:
- Important to note that the Stateflow placeholder function is currently set to return 0 until functionality is figured out.
- The CalculateActualSlip function has potential "Div by Zero" error if both front wheel speeds are 0.
- Please feel free to let me know whether naming conventions should be fixed for clarity.

### DIFF
--- a/firmware/shared/controls/tc_scale_factor.h
+++ b/firmware/shared/controls/tc_scale_factor.h
@@ -16,19 +16,14 @@ T CalculateActualSlip(T left_rear_wheel_speed, T right_rear_wheel_speed,
     }
 
     if (actual_slip < 0) {
-        return 0;
+        actual_slip = 0;
     }
     return actual_slip;
 }
 
 template <typename T>
-T TCStateflowChart(T actual_slip, T target_slip) {
+T CalculateTCScaleFactor(T actual_slip, T target_slip) {
     // Stateflow: Multi-Stage TC
     return 0;
-}
-
-template <typename T>
-T CalculateTCScaleFactor(T TC_scale_factor_chart_out) {
-    return TC_scale_factor_chart_out;
 }
 }  // namespace ctrl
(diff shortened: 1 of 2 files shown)


firmware/shared/controls/tc_scale_factor.h
```diff
@@ -0,0 +1,34 @@
...
+    T actual_slip;
+
+    if (left_rear_wheel_speed > right_rear_wheel_speed) {
+        actual_slip = (left_rear_wheel_speed / idle_wheel_spd) - 1;
+    } else {
+        actual_slip = (right_rear_wheel_speed / idle_wheel_spd) - 1;
+    }
+
+    if (actual_slip < 0) {
+        return 0;
```
BlakeFreer, line 19: since we are "limiting" the actual slip to be positive, it is more clear to set `actual_slip=0` in this `if` then have a single return

firmware/shared/controls/tc_scale_factor.h
```diff
@@ -0,0 +1,34 @@
...
+}
+
+template <typename T>
+T TCStateflowChart(T actual_slip, T target_slip) {
+    // Stateflow: Multi-Stage TC
+    return 0;
+}
+
+template <typename T>
+T CalculateTCScaleFactor(T TC_scale_factor_chart_out) {
```
BlakeFreer, line 31: why do we have a function which does nothing?

BlakeFreer (changes requested):

firmware/shared/controls/tc_scale_factor.h
ojesene1: I was following the 3-function breakdown you explained to me, but I see now that it's unnecessary in this case since the StateFlow chart produces our final value.

BlakeFreer (approved):

PR #302, opened 2024-11-08
Motor Torque logic and testing

jabrap1:
Need feedback on how to incorporate finite state machine into final code.
Additionally, we are unsure of how to test the finite state machine for multiple states, we have tested for starting in Run State, but cannot test for it without changing the original code (finite state machine). The code CAN be updated in order to pass its past state in as a parameter, but we dont know how the team wants to tackle finite state machine yet.

### DIFF
--- a/firmware/shared/controls/motor_torque.h
+++ b/firmware/shared/controls/motor_torque.h
@@ -1,8 +1,7 @@
 #pragma once
-#include <stdbool.h>
-
-#include <queue>
 #include <tuple>
+
+#include "../util/moving_average.h"
 
 // Peter Jabra and Aleeza Ali Zar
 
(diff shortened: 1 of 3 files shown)


firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,99 @@
...
+        std::swap(samples, empty);
+    }
+
+private:
+    std::queue<T> samples;
+    T sum;
+};
+
+template <typename T>
+std::tuple<T, T> CalculateMotorTorque(T raw_torque_values[], int size,
```
BlakeFreer, line 47: input should be a single value, not an array

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,99 @@
...
+
+        right_motor_torque_limit = running_average * right_factor;
+        left_motor_torque_limit = running_average * left_factor;
+    }
+
+    return std::tuple(right_motor_torque_limit, left_motor_torque_limit);
+}
+
+template <typename T>
+bool CheckBrakePedalPosition(T brake_pedal_position) {
```
BlakeFreer, line 65: inline this function.

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,99 @@
...
+    return brake_pedal_position > static_cast<T>(10);
+}
+
+template <typename T>
+T ComputeTorqueRequest(T driver_torque_request, T brake_pedal_position) {
+    State currentState = State::Stop;
```
BlakeFreer, line 71: static to hold state

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,99 @@
...
+    bool brake_on = CheckBrakePedalPosition<T>(brake_pedal_position);
+
+    T motor_torque_request;
+
+    while (true) {
```
BlakeFreer, line 76: can't have this loop in a function since that will block the program from continuing, you can assume it is called elsewhere in a loop

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,99 @@
...
+        switch (currentState) {
+            case State::Run:
+                motor_torque_request = driver_torque_request;
```
BlakeFreer, line 79: replace all of this code with 2 `if/elseif` statements - the first updates `current_state =` based on the pedal states and the second chooses which value to return

firmware/shared/controls/motor_torque.h
BlakeFreer: also use `snake_case`

firmware/shared/controls/motor_torque_test.cc
```diff
@@ -0,0 +1,80 @@
...
+#include "testing.h"
+
+int main() {
+    using namespace ctrl;
+
+    // Test cases for CalculateMotorTorque
+    // Test Case 1: Smaller Sample Sizes
+    {
+        double raw_torque_values[] = {5, 10, 15, 20};
+        auto result = CalculateMotorTorque(raw_torque_values, 4, 1.5, 1.2);
```
BlakeFreer, line 14: Use "structures bindings" to unpack the tuple into separate variables here rather than with the get method 

https://en.cppreference.com/w/cpp/language/structured_binding

BlakeFreer (changes requested):

BlakeFreer (changes requested):
move away from a class and add a reset flag to CalculateMotorTorque

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,63 @@
...
+        current_state = State::Run;
+    }
+
+    switch (current_state) {
+        case State::Run:
+            return driver_torque_request;
+
+        case State::Stop:
+            return static_cast<T>(0.0);
+        default:
```
BlakeFreer, line 58: since we are switching based on an `enum class`, it is impossible for the value to not be Start or Stop, so we shouldn't have a `default` case

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,63 @@
...
+    static shared::util::MovingAverage<T, 10> running_average;
+
+    if (reset) {
+        running_average = shared::util::MovingAverage<T, 10>();
+    }
+
+    running_average.LoadValue(new_torque_value);
+
+    T running_average_value = running_average.GetValue();
+    T scaled_running_average = running_average_value * 10;
```
BlakeFreer, line 27: what is this `*10` for?

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,63 @@
...
+};
+
+namespace ctrl {
+
+template <typename T>
+std::tuple<T, T> CalculateMotorTorque(T new_torque_value, T right_factor,
+                                      T left_factor, bool reset = false) {
+    static shared::util::MovingAverage<T, 10> running_average;
+
+    if (reset) {
```
BlakeFreer, line 20: good

firmware/shared/controls/motor_torque.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+#include <tuple>
+
+#include "shared/util/moving_average.h"
+
+// Peter Jabra and Aleeza Ali Zar
+
+enum class State {
```
BlakeFreer, line 8: move this enum inside the `ctrl` namespaec

firmware/shared/controls/motor_torque_test.cc
```diff
@@ -0,0 +1,91 @@
...
+
+int main() {
+    using namespace ctrl;
+
+    // Test cases for CalculateMotorTorque
+    // Test Case 1: Smaller Sample Sizes
+    {
+        double new_torque_values[] = {5.0, 10.0, 15.0};
+        for (int i = 0; i < 3; i++) {
+            CalculateMotorTorque(new_torque_values[i], 1.5, 1.2, i == 0);
```
BlakeFreer, line 15: nice use of `i` to reset

firmware/shared/controls/motor_torque_test.cc
```diff
@@ -0,0 +1,91 @@
...
+        double new_torque_value = 10.0;
+        auto [right_motor_torque, left_motor_torque] =
+            CalculateMotorTorque(new_torque_value, 1.5, 1.2, true);
+        ASSERT_CLOSE(right_motor_torque, 150.0);
+        ASSERT_CLOSE(left_motor_torque, 120.0);
+    }
+
+    // Test cases for ComputeTorqueRequest
+    ASSERT_CLOSE(ComputeTorqueRequest(30.0, 20.0),
+                 0.0);  // Should return 0.0 due to State::Stop
```
BlakeFreer, line 81: add two asserts between these tests where the driver_torque_req and brake values go below their threshold individually. We want to make sure we stay in the `::Stop` state until both happen simultaneously

BlakeFreer (changes requested):

firmware/shared/controls/motor_torque.h
jabrap1: So the numbers passed into this function will be inherited from previous functions which return a number from 1-100, however, we need a output a number from 1- 1000 as thats what is used the later functions. While we could change the scale to just be smaller, im not sure how sensitive later functions will be. They might require numbers between 1-1000, so for now i believe we should keep it as is, and leave a note to scale it down if we find out later functions dont require sensitive values

BlakeFreer: let's remove it from this function. it sound like the scale factor is more of an interface between the functions, so we can do the math there. i think it muddles the behaviour of this one

firmware/shared/controls/motor_torque_test.cc
```diff
@@ -79,12 +79,25 @@ int main() {
     // Test cases for ComputeTorqueRequest
     ASSERT_CLOSE(ComputeTorqueRequest(30.0, 20.0),
                  0.0);  // Should return 0.0 due to State::Stop
+    ASSERT_CLOSE(ComputeTorqueRequest(15.0, 2.5),
+                 0.0);  // Only the !brake_on condition is met here, so it
+                        // should stay in State::Stop and return 0.0
+    ASSERT_CLOSE(
+        ComputeTorqueRequest(4.5, 10.5),
+        0.0);  // Only the driver_torque_request < static_cast<T>(5) condition
+               // is met here, so it should stay in State::Stop and return 0.0
```
BlakeFreer, line 88: great tests

BlakeFreer (approved):
good work!

Issue #201, opened 2024-10-02
CAN Errors


BlakeFreer:
# CAN Errors

Number of members: 2

Relevant skills: C++, CAN, Go

Timeline: Mid November

## Description

We have no mechanism for reporting errors from the vehicle microcontrollers. This project will send errors over the CAN line and process them on a host machine.

CAN messages can hold 64 bits of data. That's 64 boolean values which can each indicate whether a specific error has occured!

We should have a global "Error" object for each microcontroller which has 64 boolean values. Each project is free to define what errors it will report. For example, a FreeRTOS project should be able to indicate a Stack Overflow error, while the FrontController (which reads pedal positions) should be able to report when a sensor value is out of the expected range.

We will periodically send the Error struct over CAN. Other devices (like our data logger or Raspberry Pi) will log these messages and display them to the developer for debugging.

## Tasks

This is a two person project. One member (upper year) will write the firmware to handle and send the messages while the other (first year) will write a command line tool which displays the error statuses. The interface between the two members is a DBC file which describes what each error message means.

Work together on steps 1 and 2, then break off for step 3a and 3b (these represent the bulk of the work for this issue).

### 1. Start a new Demo Project

This project is your deliverable. Follow these instructions to set it up <https://macformula.github.io/racecar/firmware/project-structure/>

You only need to write a CLI platform layer, not stm.

### 2. Create a CAN DBC for the Errors

There should be 1 message with 64 one-bit signals. You can call the signals `error0`-`error63` for now.

There is no specified recipient for the message, so use `Vector__XXX`.

Create a `config.yaml` file to generate the C++ code from the DBC. See the `Demo/CAN` projects for an example config.

### 3a. C++ Implementation

A CAN object with 64 booleans has been created for you. Decide how you want to set and clear error bits throughout the program. In your Demo Project, mock up error states. Continuously (every 500 ms) send the CAN message.

### 3b. CLI Viewer

Create a CLI program, in Go, which reads CAN messages from a CAN socket (talk to me if you need to set up a linux environment).

You can put your code in a folder under `racecar/scripts`. Choose a cool name for your program (not just `CAN_Error_CLI`).

Here are some milestones you can strive for:

1. Print incoming CAN messages to stdout.
2. Parse the DBC file to print the names of the Errors which are currently set.
3. Collect multiple instances of the same error message into one location to keep the UI clear.
4. Allow the user to "Acknowledge" an error to remove it from the screen.
5. Parse command line flags to ignore specific errors.

[...]

PR #314, opened 2024-11-10
[Issue #201] CAN errors C++ implementation

luaibash:
## Description
- Systems within the vehicle must communicate on if an error has occurred. We need an interface for errors such that we can access and throw a specific error when it has occurred, and have it be sent through CAN for other systems to determine if an error has occurred.
- Created new project CANErrors to hold this implementation.
- Added a dbc file that generates a 64 bit error integer, where each bit will be treated as a separate error, 64 total errors.
- Created new class ErrorHandler to hold this error object, and define methods to set different errors, and send them.
- Defined an enum that holds all 64 errors.


firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+// Interface to set/send all possible errors for the system
+class ErrorHandler {
+    private:
+        // Object that holds a 64 bit int, each bit representing an error
+        generated::can::TMS_ERROR error_msg{};
+    public:
+        // Sets the error based on the error index given
+        void setError(uint32_t error) {
+            if (error < 0 || error > 63) {
+                std::cerr << "Error index must be between 0 and 63!" << std::endl;
```
luaibash, line 23: Up to discussion on how the error handling should be implemented here. I just added a simple log output, but the way this error should be handled should align to how we do it for this whole project.

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+            error_msg.errors |= (1ULL << error);
+        }
+
+        // Sends the error message through the provided CAN bus
+        void sendError(shared::can::CanBus error_can_bus) {
+            error_can_bus.Send(error_msg);
+
+            // Reset after a send to not send duplicate errors
+            resetError();
+        }
```
luaibash, line 36: Made the CAN bus to send the error object to as a parameter for this function. This is okay if the bus is changing and will not always be the same one.

However if it is always the same one, this can just be set as a private class variable to be used internally by this method. Let me know thoughts on what is expected.

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+***************************************************************/
+
+// Interface to set/send all possible errors for the system
+class ErrorHandler {
+    private:
+        // Object that holds a 64 bit int, each bit representing an error
+        generated::can::TMS_ERROR error_msg{};
+    public:
+        // Sets the error based on the error index given
+        void setError(uint32_t error) {
```
BlakeFreer, line 21: This input should be your enum

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+            if (error < 0 || error > 63) {
+                std::cerr << "Error index must be between 0 and 63!" << std::endl;
+                return;
+            }
+
+            error_msg.errors |= (1ULL << error);
+        }
+
+        // Sends the error message through the provided CAN bus
+        void sendError(shared::can::CanBus error_can_bus) {
```
BlakeFreer, line 31: Good, like this a lot

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+            error_can_bus.Send(error_msg);
+
+            // Reset after a send to not send duplicate errors
+            resetError();
+        }
+
+        // Resets all errors back to untriggered
+        void resetError() {
```
BlakeFreer, line 39: This should be private so it cannot be called directly, only by SendError

firmware/projects/Demo/CANErrors/inc/app.h
BlakeFreer: The error message is only defined for a specific bus (it's only present in one dbc) and doesn't make sense to send on an arbitrary bus. We will likely generate this class so don't worry about that part of it

BlakeFreer (changes requested):

BlakeFreer:
In general, functions, variables, constants etc should be named following this guide 

https://google.github.io/styleguide/cppguide.html#Naming

firmware/projects/Demo/CANErrors/inc/app.h
luaibash: Alright, should I leave it as be or define the bus as a private member then?

firmware/projects/Demo/CANErrors/README.md
```diff
@@ -1,13 +1,36 @@
 # CANErrors
+This is the CAN Errors demo project. It defines a global system for the vehicle to send/receive errors through different subsystems. It defines:
+- A 64 bit CAN error signal. Each bit represents a different error, and will be sent periodically for other systems to know if an error has occurred
+- The `ErrorHandler` class, an interface to set/send different errors on the CAN signal
+- Enum `Error` that defines all 64 possible errors to send through the CAN signal
 
-This is the CAN Errors demo project. This defines the mechanism for reporting errors from the vehicle microcontrollers. This project will send errors over the CAN line and process them on a host machine.
+<br>
```
BlakeFreer, line 7: this is a bit pedantic but stick to the regular markdown formatting i.e. accept the default paragraph spacing instead of adding a `<br>`

firmware/projects/Demo/CANErrors/README.md
```diff
@@ -1,13 +1,36 @@
...
+### Sending errors with ErrorHandler:
+- `sendError` takes a CAN bus and sends the errors through that bus. It resets the errors to send the next set of errors
+```
+error_handler.sendError(error_can_bus);
+```
+
+<br>
 
 ## Building the Project
 - To build the project, run the following in the firmware directory "make PROJECT=Demo/CANErrors PLATFORM=cli"
```
BlakeFreer, line 33: use backticks \` \` around the command to format it like code

BlakeFreer: also for the `./build` cmd

firmware/projects/Demo/CANErrors/errors.dbc
```diff
@@ -0,0 +1,9 @@
+VERSION ""
+
+BS_:
+
+BU_: TMS
+
+// Defines one 64 bit signal where each bit represents a different error
+BO_ 390 TMS_ERROR: 8 TMS
+    SG_ Errors : 0|64@1- (1,0) [0|0] "bitfield" Vector__XXX
```
BlakeFreer, line 9: this generates an `int64_t` but we're using it for the bit value so `uint64_t` is better. Check the DBC syntax for how to specify an unsigned value

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include "../generated/can/error_can_messages.h"
+#include "shared/comms/can/can_bus.h"
+
+/***************************************************************
+    App-level objects
+***************************************************************/
+
+// Defines all possible errors to set
+enum Error {
```
BlakeFreer, line 15: this should be an `enum class`

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+    Error58 = 58,
+    Error59 = 59,
+    Error60 = 60,
+    Error61 = 61,
+    Error62 = 62,
+    Error63 = 63
+};
+
+// Interface to set/send all possible errors for the system
+class ErrorHandler {
```
BlakeFreer, line 83: follow https://google.github.io/styleguide/cppguide.html#Naming for naming conventions. see instance variables & function names

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+        generated::can::TMS_ERROR error_message{};
+
+        // Resets all errors back to untriggered
+        void resetError() {
+            error_message.errors = 0;
+        }
+    public:
+        // Sets the error based on the error index given
+        void setError(Error error) {
+            if (error < 0 || error > 63) {
```
BlakeFreer, line 95: after `Error` is an enum class, you do not need this check since you can assume there are no invalid enum values.

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+                std::cerr << "Error index must be between 0 and 63!" << std::endl;
+                return;
+            }
+
+            error_message.errors |= (1ULL << error);
```
BlakeFreer, line 100: after making this an enum class, you will need to cast back to int

firmware/projects/Demo/CANErrors/inc/app.h
```diff
@@ -0,0 +1,110 @@
...
+    Error59 = 59,
+    Error60 = 60,
+    Error61 = 61,
+    Error62 = 62,
+    Error63 = 63
+};
+
+// Interface to set/send all possible errors for the system
+class ErrorHandler {
+    private:
```
BlakeFreer, line 84: public section should go before private section since it holds the function's exposed API. See https://google.github.io/styleguide/cppguide.html#Declaration_Order

BlakeFreer (changes requested):

firmware/projects/Demo/CANErrors/inc/app.h
BlakeFreer: Just add a comment in the code for now

BlakeFreer (approved):

Issue #293, opened 2024-11-05
cangen: architecture & function responsibility


BlakeFreer:
Some things I don't like about the current architecture:

- ~~the `if sig.byte_order` block in `_get_masks_shifts()` is pretty repetitive. We're really just mapping a byte order string to a function. This can be more direct with a dictionary with string keys and function values.~~ the current code is more obvious, don't change this
  - cantools also guarantees that the byteorder is valid so we don't need an explicit check
- we create the Jinja environment multiple times (twice per bus). This should be abstracted out and only run once. Maybe it gets made in `generate_can_from_dbc` and passed into `generate_code`? Would be nice to have a function that creates and returns the environment with all filters set
- after the last bullet, is there still value in the `generate_from_jinja_template` function? Should it be inlined into `generate_code`? It's a bit of a strange function given how many parameters it requires and how little it does. Maybe the two calls to this function could be replaced by a loop over `(CAN_MESSAGES_TEMPLATE_FILENAME, MSG_REGISTRY_TEMPLATE_FILENAME)`or something simila
- i'm not a huge fan of variables/ return types that look like `Dict[str, Dict[str, Tuple[List[int], List[int]]]]:` This is really hard to understand. We should at least use dataclasses to bundle some of this information into a named type.
- Log levels: since `cangen` can be used standalone, we shouldn't require the user to know cmake logging levels to use it. We should have two mutually exclusive arguments: a `--verbose` aka `-v` flag and `--cmake-log-level=###` option, the latter of which replaces the current `--log-level`. The `parse()` function should figure out which was passed and convert that to a python logging level. You can then set `args.loglevel` to this value and directly pass `args.level` to `logger.setLevel
  - this will require you to store and manipulate the value from `parser.parse_args()` before returning it
- the current functions names aren't great. Renaming them may also help decide how to organize the behaviour
  - if you rename the cangen entry point, make sure to update `pyproject.toml`.

PR #315, opened 2024-11-11
andrewi/293 cangen refactor

AndrewI26:
Refactored cangen as specified in #293

### DIFF
--- a/scripts/cangen/cangen/can_generator.py
+++ b/scripts/cangen/cangen/can_generator.py
@@ -9,7 +9,7 @@ import os
 import re
 import shutil
 import time
-from typing import Dict, List, Tuple, Callable, Any
+from typing import Dict, List, Tuple
 
 import numpy as np
 from cantools.database import Database, Message, Signal
@@ -22,12 +22,8 @@ logger = logging.getLogger(__name__)
 EIGHT_BITS = 8
 EIGHT_BYTES = 8
 TOTAL_BITS = EIGHT_BITS * EIGHT_BYTES
-MSG_REGISTRY_FILE_NAME = "_msg_registry.h"
-CAN_MESSAGES_FILE_NAME = "_can_messages.h"
 
 TEMPLATE_FILE_NAMES = ["can_messages.h.jinja2", "msg_registry.h.jinja2"]
-
-PACKAGE_NAME = "cangen"
 
 
 def _parse_dbc_files(dbc_file: str) -> Database:
@@ -172,33 +168,11 @@ def _camel_to_snake(text):
     return re.sub("([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
 
 
-def _create_jninja_environment(
-    package_name: str, filters: dict[str, Callable[..., Any]]
-):
-    env = Environment(
-        loader=PackageLoader(package_name), trim_blocks=True, lstrip_blocks=True
-    )
-    for filter_name, filter_function in filters.items():
-        env.filters[filter_name] = filter_function
-
-    return env
+def _create_output_file_name(bus_name: str, template_file_name: str) -> str:
+    return os.path.join(bus_name.lower() + "_" + template_file_name.removesuffix(".jninja2"))
 
 
-def _create_output_file_name(output_dir: str, bus_name: str, template_file_name: str):
-    return os.path.join(output_dir, bus_name.lower() + "_" + template_file_name[:-7])
-
-
-def _generate_jninja_template(template_file_name, output_path, context, env):
-    template = env.get_template(template_file_name)
-    rendered_code = template.render(**context)
-
-    output_file_name = output_path
-    with open(output_file_name, "w") as output_file:
-        output_file.write(rendered_code)
-    logger.info(f"Rendered code written to '{os.path.abspath(output_file_name)}'")
-
-
-def _generate_code(bus: Bus, config: Config):
+def _generate_code(bus: Bus):
     """
     Parses DBC files, extracts information, and generates code using Jinja2
     templates.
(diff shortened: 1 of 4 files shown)


scripts/cangen/cangen/main.py
```diff
@@ -16,27 +16,41 @@ def parse():
     parser = argparse.ArgumentParser(description="DBC to C code generator")
     parser.add_argument("project", type=str, help="Name of the project")
     parser.add_argument(
```
BlakeFreer, line 18: arguments look good. we shouldn't let the user pass both `--verbose` and `--cmake-log-level`, so make a mutually exclusive group for them https://docs.python.org/3/library/argparse.html#mutual-exclusion

scripts/cangen/cangen/main.py
```diff
@@ -16,27 +16,41 @@ def parse():
...
-        "--log-level",
+        "-v",
+        "--verbose",
+        dest="is_verbose",
+        action="store_true",
+        help="Enable verbose output",
+    )
+    parser.add_argument(
+        "--cmake-log-level",
```
BlakeFreer, line 26: update the variable name in `firmware/cmake/cangen.cmake`

scripts/cangen/cangen/can_generator.py
```diff
@@ -25,8 +25,9 @@
 MSG_REGISTRY_FILE_NAME = "_msg_registry.h"
 CAN_MESSAGES_FILE_NAME = "_can_messages.h"
 
-CAN_MESSAGES_TEMPLATE_FILENAME = "can_messages.h.jinja2"
-MSG_REGISTRY_TEMPLATE_FILENAME = "msg_registry.h.jinja2"
+TEMPLATE_FILE_NAMES = ["can_messages.h.jinja2", "msg_registry.h.jinja2"]
+
+PACKAGE_NAME = "cangen"
```
BlakeFreer, line 30: use the `__package__` special variable. It holds the string name of the current package. Then you don't need a constant, just use it inline

scripts/cangen/cangen/can_generator.py
```diff
@@ -25,8 +25,9 @@
 MSG_REGISTRY_FILE_NAME = "_msg_registry.h"
```
BlakeFreer, line 25: these two constants are unused now, remove

scripts/cangen/cangen/can_generator.py
```diff
@@ -171,29 +172,33 @@ def _camel_to_snake(text):
...
+    return env
 
-    # Load and render template
-    template = env.get_template(template_path)
-    rendered_code = template.render(**context_dict)
 
-    with open(output_path, "w") as output_file:
-        output_file.write(rendered_code)
+def _create_output_file_name(output_dir: str, bus_name: str, template_file_name: str):
+    return os.path.join(output_dir, bus_name.lower() + "_" + template_file_name[:-7])
```
BlakeFreer, line 188: use `removesuffix` to directly remove `.jinja2` instead of a slicing operation

scripts/cangen/cangen/can_generator.py
```diff
@@ -220,39 +225,45 @@ def generate_code(bus: Bus, config: Config):
...
 
 
-def generate_can_from_dbc(project_folder_name: str):
+def generate_can_for_project(project_folder_name: str):
+    """Generates C code for a given project.
+    Ensure the project folder contains a config.yaml file which specifies the relative path to its corresponding dbc file."""
     os.chdir(project_folder_name)
     config = Config.from_yaml("config.yaml")
 
     _prepare_output_directory(config.output_dir)
```
BlakeFreer, line 267: now that the folder is ready, lets `chdir` into it here. There we don't have to worry about prepending every file path with `config.output_dir` anymore. This actually means that we don't need to pass `config` into `_generate_code()` anymore!

scripts/cangen/cangen/can_generator.py
```diff
@@ -220,39 +225,45 @@ def generate_code(bus: Bus, config: Config):
...
-        context,
-    )
+    logger.debug("Generating code for can messages and msg registry.")
 
-    logger.debug("Generating code for msg registry")
-    _generate_from_jinja2_template(
-        MSG_REGISTRY_TEMPLATE_FILENAME,
-        os.path.join(config.output_dir, bus.bus_name.lower() + MSG_REGISTRY_FILE_NAME),
-        context,
+    env = _create_jninja_environment(
```
BlakeFreer, line 230: I'm going to walk back on what I said earlier about making the env in a function. The function has a fairly long call signature and doesn't help much with the structure. As a reader, I think it's more work to follow this function to it's definition that it would be to just see the environment get made here. Let's get rid of the `_create_jinja_env` and directly run the code here.

For the filters, I like the idea of iterating over a dictionary, but since we have so few elements I think we should just set them directly

```
env.filters["hex"] = hex
env.filters[... = _camel...
```

scripts/cangen/cangen/can_generator.py
```diff
@@ -220,39 +225,45 @@ def generate_code(bus: Bus, config: Config):
...
+        PACKAGE_NAME, {"camel_to_snake": _camel_to_snake, "decimal_to_hex": hex}
     )
+    for template_file_name in TEMPLATE_FILE_NAMES:
+        _generate_jninja_template(
```
BlakeFreer, line 234: as with the previous comment, there is so much information to pass into the function that I think we just inline the code.

Also if you chdir into `config.output_dir` (see comment in `generate_code`) then you can modify `_create_output_file_name` to only take the bus and template names

scripts/cangen/cangen/can_generator.py
```diff
@@ -172,33 +168,12 @@ def _camel_to_snake(text):
...
 
 
-def _create_jninja_environment(
-    package_name: str, filters: dict[str, Callable[..., Any]]
-):
-    env = Environment(
-        loader=PackageLoader(package_name), trim_blocks=True, lstrip_blocks=True
+def _create_output_file_name(output_dir: str, bus_name: str, template_file_name: str) -> str:
+    return os.path.join(
+        output_dir, bus_name.lower() + "_" + template_file_name.removesuffix(".jninja2")
```
BlakeFreer, line 173: typo, its `jinja` not `jninja`

BlakeFreer (changes requested):

PR #316, opened 2024-11-12
201 can errors cli viewer

sokosam:


scripts/can_errgo/Demos/listenerDemo.go
```diff
@@ -0,0 +1,92 @@
...
+	"net"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	canInterface = "vcan0"  // Use the virtual CAN interface
+	canFrameSize = 16       // CAN frame size (for a standard CAN frame)
+)
+
+func IntPow(n, m int) int {
```
BlakeFreer, line 16: isn't there a built in function for this?

scripts/can_errgo/components/table/table.go
```diff
@@ -0,0 +1,489 @@
...
+	for _, opt := range opts {
+		opt(&m)
+	}
+
+	m.UpdateViewport()
+
+	return m
+}
+
+// WithColumns sets the table columns (headers).
```
BlakeFreer, line 155: the function names explain these comments, I don't think they're necessary

scripts/can_errgo/components/table/table.go
```diff
@@ -0,0 +1,489 @@
...
+	row := lipgloss.JoinHorizontal(lipgloss.Top, s...)
+
+	if r == m.cursor && m.ShowCursor {
+		return m.styles.Selected.Render(row)
+	}
+
+	return row
+}
+
+func max(a, b int) int {
```
BlakeFreer, line 471: there should be built in functions for max min and maybe clamp

scripts/can_errgo/README.md
BlakeFreer, line 1: use markdown formatting for headers and code blocks

https://www.markdownguide.org/cheat-sheet/

scripts/can_errgo/components/box/text-box.go
BlakeFreer, line 1: small detail, but do `text-box` and `table` need to be in separate folders or can they go directly under `components/`?

scripts/can_errgo/components/table/table.go
```diff
@@ -0,0 +1,489 @@
...
+
+// Update is the Bubble Tea update loop.
+func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
+	if !m.focus {
+		return m, nil
+	}
+
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		switch {
```
BlakeFreer, line 212: is there a way to attach the key behaviour directly to the key? Maybe something like adding a `callback` function when you create the DefaultKeyMap. If not this is fine

scripts/can_errgo/Demos/listenerDemo.go
BlakeFreer: looks like you don't use this function anywhere. it can be removed

scripts/can_errgo/Demos/threadDemo.go
BlakeFreer, line 1: can this file be deleted?

scripts/can_errgo/Demos/listenerDemo.go
BlakeFreer, line 1: can this file be deleted?

scripts/can_errgo/components/table/table.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+}
+
+// SetStyles sets the table styles.
+func (m *Model) SetStyles(s Styles) {
+	m.styles = s
+	m.UpdateViewport()
+}
+
+// Option is used to set options in New. For example:
```
BlakeFreer, line 129: don't need this comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+	"github.com/charmbracelet/bubbles/key"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+	"golang.org/x/sys/unix"
+)
+
+
+const (
+	canInterface = "vcan0"  // Use the virtual CAN interface
+	canFrameSize = 16       // CAN frame size (for a standard CAN frame)
```
BlakeFreer, line 27: why is this 16? Can frames have 8 data bytes

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+
+	"github.com/charmbracelet/bubbles/key"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+	"golang.org/x/sys/unix"
+)
+
+
+const (
+	canInterface = "vcan0"  // Use the virtual CAN interface
```
BlakeFreer, line 26: the interface should be a runtime cli argument

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+	canFrameSize = 16       // CAN frame size (for a standard CAN frame)
+)
+
+// Might have future use
```
BlakeFreer, line 30: remove this function if we don't need it

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+        copy(newRows[i], rows[i])
+    }
+    return newRows
+}
+
+// Converts a 8-byte slice to a uint64, this is for masking purposes.
+func bytesToUint64(b []byte) uint64 {
+    var result uint64
+    for i := 0; i < len(b) && i < 8; i++ {
+        // Reverse the order: least significant byte (rightmost) gets the lowest bits
```
BlakeFreer, line 44: good comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+}
+
+func tickEvery(t time.Duration) tea.Cmd {
+	return func() tea.Msg {
+		time.Sleep(t)
+		return TickMsg{}
+	}
+}
+
+// Might have future use
```
BlakeFreer, line 117: remove if we don't need

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+    return -1
+}
+
+// Check if the cursor is over bounds
+func overBounds(table table.Model) bool{
+	return len(table.Rows())==0 || table.Cursor() < 0 || table.Cursor() >= len(table.Rows()) 
+}
+
+
+func (m *model) re_fresh(){
```
BlakeFreer, line 143: should be `refresh` with no underscore

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+	rows := m.table.Rows()
+	for i:=0; i < len(rows); i++{
+		if val,err := strconv.Atoi(rows[i][2]); err != nil{
```
BlakeFreer, line 146: the "freshness" should be stored with the error message struct, not in the table. It is a "code smell" to see this value being stored in the table text and converted to & from string

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+			fmt.Printf("Something went wrong during refresh!")
+		} else{
+			rows[i][2] = strconv.Itoa(val + 1)
+		}
+	}
+}
+
+// Sort the table by the specified column index.
+func (m *model) SortByColumn(columnIndex int) error {
```
BlakeFreer, line 155: Sorting should happen before the rows are placed in the table - see file comment

scripts/can_errgo/error_tui.go
BlakeFreer, line 1: You should have a list / array / vector of all of the errors in the system. Each should have a freshness value, count, etc. You should update the fields of these structs whenever a new message arrives, then sort the messages and place them in the table.

You should never be reading from the table. It should only be used to display information, not store it. You should never rely on `string.Atoi` to get values back from the table - if you need to store a value, place it in a struct and display it in the table

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+func (m model) Init() tea.Cmd { return tickEvery(m.t) }
+
+
+// Table and ignore subtable update loop.
+func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	var cmd tea.Cmd
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		if m.submenuActive{
+			// If submenu is active, only let the submenu handle KeyMsg
```
BlakeFreer, line 217: why doesn't submenu use a KeyMap like the main menu does?

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+				if val, ok := m.hiddenCounts["error" + errorNumberStr]; ok{
+					m.hiddenCounts["error" + errorNumberStr ] = val + 1
+					
+				} else{
+					m.hiddenCounts["error" + errorNumberStr] = 1
+					m.errorToBit["error" + errorNumberStr] = k
+				}
+
+				// If the error is not in the table, add it, otherwise update the count.
+				index := findRowString(m.table.Rows(),  "error" + errorNumberStr)
```
BlakeFreer, line 355: see file comment. Data should not be stored in the table

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,668 @@
...
+		
+		msg := CANMsg{ ID : canID, Value: data}
+		p.Send(msg)
+	}
+}
+
+
+func main() {
+	// Define the command line flag
+	warnFlag := flag.String("w", "", "warning time for the table (integer value)")
```
BlakeFreer, line 632: The purpose of the flag package is to handle default arguments and type parsing. You should be able to remove the whole `if warnFlag else` block

BlakeFreer (changes requested):

BlakeFreer:
The table rendering and program loop code looks good. There's a few Demo files that shouldn't be included in the merge. Use git rm to delete them.

I noticed that information about the error messages (like freshness) is stored in the table itself. This isn't the best design. The table should only be used as an output to represent the internal state of the system. You should have an array / list or similar holding a struct for each message type that includes all relevant fields, then sort that array by freshness and display those structs in the table rows. This will remove functions like re_fresh, SortByColumn, and FindRowString which operate on the table data. After this fix is made, you should (almost) never need to call string.Atoi on any table values - this data should be stored as an integer somewhere else.

scripts/can_errgo/Demos/listenerDemo.go
sokosam: I believe there is an built in function for float64. Anyhow, I have removed the function.

scripts/can_errgo/components/box/text-box.go
sokosam: They can but it could cause ambiguity during import as go does not allow single-file imports. I have changed it to have both be under components in the most recent commit.

scripts/can_errgo/components/table/table.go
sokosam: These functions were left by the original creator, though I will note the built in max/min functions takes in float64 as parameters and returns float64. Whereas our function deals with int which is appropriate since we do not expect anything other than natural numbers.

scripts/can_errgo/error_tui.go
sokosam: Removed

scripts/can_errgo/error_tui.go
sokosam: Refactored

scripts/can_errgo/error_tui.go
sokosam: Although the data frames is only 8 bytes the vcan has the structure:

4 bytes for Identifier
1 byte for DLC
1 byte padding
2 byte reserved
8 byte payload

I assume we will need the other bytes for information (Error name and description)

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,626 @@
...
+func RowValuesToRow(r RowValues) table.Row{
+	return table.Row{r.Error, strconv.Itoa(r.Count), strconv.Itoa(r.Recency)}
+}
+
+// Convert a slice of RowValues to a slice of table.Rows
+func toTableRows(r []RowValues) []table.Row {
+	var rows []table.Row
+	for _, v := range r {
+		// Check if the RowValues is "non-empty" and Active
+		if v.Error != "" && v.Count != 0 && !v.Ignored{
```
BlakeFreer, line 51: When will we ever have `.Error = ""`? Is this possible? What does it mean?

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,626 @@
...
+			m.rowsValues[i].Count = 0
+			m.rowsValues[i].Recency = 0
+		}
+	}
+}
+
+// getIgnoredRows returns a slice of table.Rows that are ignored and have an error message.
+func (m model) getIgnoredRows() []table.Row {
+	var rows []table.Row
+	for i := range(len(m.rowsValues)){
```
BlakeFreer, line 117: can you iterate over the index and rowValue together so you don't need to constant access `m.rowsValues[i]`?

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,626 @@
...
+	out = out + baseStyle.Render(m.table.View()) + "\n"
+
+	// Render the submenu if it is active
+	if m.submenuActive{
+		out = out + baseStyle.Render(m.submenuTable.View()) + "\n"
+	}
+
+	// Render the box 
+	if m.submenuActive && m.submenuTable.SelectedRow() != nil{
+		m.box.SetText(m.submenuTable.SelectedRow()[0], m.getDescription(m.submenuTable.SelectedRow()[0]))
```
BlakeFreer, line 448: This section is reading data from the table. You should have a separate variable to store the index of the currently selected row and use this to directly get the description for that error struct. You also shouldn't be searching for data based on the string name. This should be an index representing the error number.

I think it would be good to rename the `RowValues` struct to `ErrorData`. You should have an array of 64 `ErrorData` structs each with their own name, description, count, recency, etc. When a new error message arrives, update the counts & recencies in this array, but _don't sort this array_. Keep it in order so you always know that error 7 is at position 7. You have some functions which iterate over the table for information (like `getDescription`, `findError`, `resetRowValue`, `getIgnoredRows`) which should all be direct lookups in the ErrorData array.

For example, you should have something in your program which is structured like

```
errors = [
    ErrorData(name="Error0", desc="this is desc of 0", count = 7, recency = 1, ignored=false),
    ErrorData(name="Error1", desc="this is desc of 1", count = 5, recency = 2, ignored=true),
    ErrorData(name="Error2", desc="this is desc of 2", count = 1, recency = 0, ignored=false),
    ErrorData(name="Error3", desc="this is desc of 3", count = 10, recency = 3, ignored=true),
    .... for all 64 errors
]
```

of course this array would be created at runtime, not hardcoded. This should happen right when the program boots.

Then `getDescription` would just be `errors[index].description`.

When a new message comes in, to update the count & recency:

```
for i, msg in range errors:
    msg.recency += 1
    if can_error bit i is set
        msg.recency = 0
        msg.count += 1
```

You will need to figure out how to sort the errors for displaying in the table, but this should be considered after getting the backend structure correct. Think about your program as a whole system with two responsibilities

1. Track the state of each message
2. Sort and display some of the messages in a table

Your logic for 1) should not rely on how 2) works. let's do 1) really well before worrying about how exactly 2) will function

BlakeFreer (changes requested):

scripts/can_errgo/README.md
BlakeFreer, line 1: This README looks really good now. The formatting is professional and clearly explains what the user needs to know. Only thing to fix is the extra spaces
- After the "(" in the "Users can" section: i.e. should be "(Hide" not "( Hide"
- Remove the double spaces in the commands ex change `sudo  apt-get` to `sudo apt-get`

scripts/can_errgo/components/table.go
```diff
@@ -13,18 +12,25 @@ import (
...
+}
+
 // Model defines a state for the table widget.
 type Model struct {
 	KeyMap KeyMap
 	Help   help.Model
 
 	cols   []Column
-	rows   []Row
+	metaRow []MetaRow
```
BlakeFreer, line 27: this is an array so it should be a plural word "metaRows" instead of "metaRow"

scripts/can_errgo/error_tui.go
```diff
@@ -38,96 +38,72 @@ var baseStyle = lipgloss.NewStyle().
...
 		}
 	}
 	return rows
 }
 
-// SortRowValuesBy sorts a slice of RowValues by the specified column, this creates a deep copy of the input slice.
-func SortRowValuesBy(r []RowValues, column int) []RowValues {
+// SortRowValuesByRecency sorts a slice of ErrorData by the Recency column.
+// This creates a deep copy of the input slice.
+func SortRowValuesByRecency(r []ErrorData) []ErrorData {
```
BlakeFreer, line 60: this looks much better

scripts/can_errgo/error_tui.go
```diff
@@ -315,31 +264,24 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-			currentError := m.table.SelectedRow()
-			newRowValues :=  SortRowValuesBy(m.rowsValues,2)
-			newRows := toTableRows(newRowValues)
-			m.table.SetRows(newRows)
-
-			// Fix the cursor back onto the row it was on before the sort.
-			if currentError != nil{
-				newIndex := findError(currentError[0], newRowValues)
-				m.table.SetCursorAndViewport(newIndex)
+		for i,errorVal := range m.errorData{
```
BlakeFreer, line 267: this is much better

scripts/can_errgo/error_tui.go
```diff
@@ -460,6 +399,74 @@ func initViewer(warning int) model{
 	mainTable := createTable()
 	table2 := createSubMenu()
 
+	// This can be determined at runtime later (DBC file parsing), for now it is hardcoded.
+	customErrors := []ErrorData{
```
BlakeFreer, line 403: create this with a loop

BlakeFreer (changes requested):
Very good improvements. Few details but overall looks good

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+	submenuKeys   KeyMap
+
+	// Slice of row information and ignore mask for incoming errors/bits
+	errorData  []ErrorData
+    ignoreMask   uint64
+
+	// Timeout flag and last CAN message time
+	isTimeout   bool
+	lastCANTime time.Time
+	t           time.Duration
```
BlakeFreer, line 135: this variable should be renamed to something more descriptive like `.timeout`

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+	return rows
+}
+
+type ErrorData struct{
+	Error string
+	Count int
+	Recency int
+	Description string
+	Ignored bool
+	//  This is used for keeping track of the error bit after sorting
```
BlakeFreer, line 105: good comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+	lastCANTime time.Time
+	t           time.Duration
+
+	// Box for displaying error descriptions
+	box box.Box
+	showBox bool
+}
+
+
+// toggleBit toggles the specified bit in the number
```
BlakeFreer, line 143: remove comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+func tickEvery(t time.Duration) tea.Cmd {
+	return func() tea.Msg {
+		time.Sleep(t)
+		return TickMsg{}
+	}
+}
+
+func (m model) Init() tea.Cmd { return tickEvery(m.t) }
+
+// Table and ignore subtable update loop.
```
BlakeFreer, line 159: remove comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	var cmd tea.Cmd
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		if m.submenuActive{
+			// If submenu is active, only let the submenu handle KeyMsg
```
BlakeFreer, line 165: good comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+				
+				// Toggle the mask for the selected error and delete the row. 
+				errorIndex := m.submenuTable.SelectedIndex()
+				m.ignoreMask = toggleBit(m.ignoreMask, errorIndex)
+				m.errorData[errorIndex].Ignored = false
+				newRows := m.getIgnoredRows()
+				m.submenuTable.SetRows(newRows)
+
+			case key.Matches(msg, m.submenuKeys.s):
+				m.submenuActive = false  // Hide submenu
```
BlakeFreer, line 186: don't need these 4 comments

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+				m.errorData[i].Recency = 0
+			}
+		}
+		// Sort the table by the recency.
+		currentError := m.table.SelectedIndex()
+		newRowValues :=  SortRowValuesByRecency(m.errorData)
+		newRows := toTableRows(newRowValues)
+		m.table.SetRows(newRows)
+
+		// Fix the cursor back onto the row it was on before the sort.
```
BlakeFreer, line 280: good comment

scripts/can_errgo/error_tui.go
```diff
@@ -0,0 +1,633 @@
...
+	// If the message is a tick, check if the last CAN message was received more than the timeout duration ago
+	case TickMsg:
+		if time.Since(m.lastCANTime) > m.t {
+			m.isTimeout = true
+		}
+		return m, tickEvery(m.t)
+	}
+
+	// Update the table with the new message
+	if _, ok := msg.(CANMsg); ok {
```
BlakeFreer, line 295: what is this condition checking? It seems to only update the main menu in one case and the submenu in the other, contrary to the next comment

BlakeFreer (commented):
some comments should be removed since they don't add any information which isn't already clear from the function title. I marked some examples of good comments and ones which can be removed

scripts/can_errgo/error_tui.go
BlakeFreer: ok thanks for the explanation

BlakeFreer (approved):

PR #322, opened 2024-11-15
Freerb/242 update can drivers

### LINKED ISSUE #242: Update CAN drivers
The current CAN peripheral and shared classes have three major drawbacks:

1. Not interrupt based - messages are not handled upon arrival, rather upon manual request.
2. Single message capacity at the bus level - old messages are lost and the latest message can be "read" multiple times. This prevents us from using CAN for one-off messages like button signals.
3. No filtering: the `stm32` platform has a hardware CAN message filter which can ignore ranges of message IDs. Without filtering, our ECUs will be bogged down handling unrelated messages.

In particular, the current implementation freezes an ECU if it does not read all messages which arrive on the STM platform. This manifested in the TMS code freezing, and the root cause was the `CAN_IT_RX_FIFO0_MSG_PENDING` interrupt being triggered by BMS->FC CAN traffic (unrelated to the TMS). Since the TMS does not need to read these messages, it never called `HAL_CAN_GetRxMessage()`, so the interrupt continuously triggered preventing the program from continuing.

My initial idea is to enable filtering for ranges of messages (perhaps autogenerated from the DBC if possible) then trigger the `HAL_CAN_GetRxMessage()` in the `CAN_IT_RX_FIFO0_MSG_PENDING` interrupt. This will place the incoming message into a queue with all other messages of the same ID. The main program can pop elements from this queue when it wants to "read" that message. Using a queue ensures that no messages are missed, and we can intelligently raise an error if the queue becomes too full (i.e. a message is being sent to us but we aren't handling it).

BlakeFreer:
closes #242


firmware/projects/TMS/inc/app.h
```diff
@@ -173,9 +173,15 @@ class BmsBroadcaster {
         bms_broadcast_.avg_therm_value = avg_thermistor_value;
         bms_broadcast_.num_therm_en = num_thermistors_;
         bms_broadcast_.therm_module_num = kThermistorModuleNumber;
+
         bms_broadcast_.checksum = CalculateBmsBroadcastChecksum(bms_broadcast_);
 
-        can_bus_.Send(bms_broadcast_);
+        auto checksum =             high_therm_id + high_therm_value +
```
TylerStAmour, line 179: Remove the calculateBmsBroadcastChecksum function and just use the inline checksum calculation

TylerStAmour (changes requested):

TylerStAmour (approved):

BlakeFreer (commented):
lg

firmware/projects/Demo/CAN/Receive/main.cc
```diff
@@ -4,27 +4,24 @@
...
+#include "shared/comms/can/bus.h"
 
-generated::can::DemobusMsgRegistry veh_can_registry{};
+using namespace generated::can;
 
-shared::can::CanBus veh_can_bus{
-    bindings::veh_can_base,
-    veh_can_registry,
-};
+DemobusBus veh_can_bus{bindings::veh_can_base};
```
samparent97, line 13: All this demo bus and veh bus is a bit odd I think you should pick one or the other.

firmware/projects/Demo/CAN/Send/main.cc
```diff
@@ -4,29 +4,29 @@
...
-shared::can::CanBus veh_can_bus{
-    bindings::veh_can_base,
-    veh_can_registry,
-};
+DemobusBus can_bus{bindings::veh_can_base};
+
+typedef struct {
+    int a;
+    int b;
+} stri;
```
samparent97, line 17: What's stri? Maybe a better name on this one.

samparent97 (commented):
Don't have my laptop with me tonight. I left a couple of quick comments, but I'll have to come back to this one in the morning.

firmware/projects/Demo/CAN/Receive/main.cc
BlakeFreer: good catch

firmware/shared/periph/can.h
```diff
@@ -5,15 +5,28 @@
...
+}  // namespace shared::can
+
 namespace shared::periph {
 
 class CanBase : public util::Peripheral {
 public:
-    virtual void Send(const shared::can::RawCanMsg&) = 0;
-    virtual void ReadQueue(shared::can::RawCanMsg[], size_t len) = 0;
+    virtual void Send(const shared::can::RawMessage&) = 0;
+    void AddToBus(can::RawMessage msg);
```
samparent97, line 21: This is awesome

firmware/mcal/cli/periph/can.h
```diff
@@ -8,44 +8,37 @@
 #include <string.h>
```
samparent97, line 8: seems like this breaks the `stm32f767` platform, I think this should be fixed before this PR goes in

firmware/projects/Demo/CAN/Send/main.cc
```diff
@@ -4,29 +4,22 @@
...
     while (true) {
-        veh_can_bus.Update();
-
-        btn_msg.state = bindings::button.Read();
-
-        veh_can_bus.Send(btn_msg);
+        TxButtonStatus msg{
+            .state = bindings::button.Read(),
+        };
+        can_bus.Send(msg.pack());
```
samparent97, line 22: Not a huge fan of having to call pack on the message when passing it into the Send function. Why not make it a template with a concept that makes sure the pack function exists?

firmware/projects/Demo/CAN/Send/main.cc
```diff
@@ -4,29 +4,22 @@
...
+DemobusBus can_bus{bindings::demo_can_base};
 
 int main(void) {
     bindings::Initialize();
     uint32_t interval_ms = 50;
 
-    generated::can::ButtonStatus btn_msg{};
-
     while (true) {
-        veh_can_bus.Update();
```
samparent97, line 25: Really nice to get rid of this

firmware/projects/Demo/CAN/Receive/platforms/cli/bindings.cc
```diff
@@ -12,17 +12,17 @@
 namespace mcal {
```
samparent97, line 12: Again, we should get this all working with the `stm32f767` platform before merging.

firmware/projects/Demo/CAN/Receive/main.cc
```diff
@@ -4,27 +4,24 @@
...
 
-    generated::can::ButtonStatus btn_msg{};
-
     while (true) {
-        veh_can_bus.Update();
-        veh_can_bus.Read(btn_msg);
+        auto btn_msg = demo_can_bus.PopRxButtonStatus();
 
-        bindings::indicator.Set(btn_msg.state);
+        if (btn_msg.has_value()) {
```
samparent97, line 21: Add a comment about what this is checking for

scripts/cangen/cangen/templates/messages.h.jinja2
```diff
@@ -33,55 +54,41 @@ enum class {{ sig.name }} {
...
-    void Unpack(const shared::can::RawCanMsg& raw_msg) override {
-        if (raw_msg.header.id != kCanId) {
-            return;
-        }
-
-        tick_timestamp = raw_msg.tick_timestamp;
-
-        // temporary signal variables
-    {% for sig in msg.signals %}
+    friend class {{ bus }};
```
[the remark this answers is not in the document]
BlakeFreer: It let's me make the Rx message constructor private and only available to the Bus. This prevents artificial messages from appearing in the system. It's all handled by autogen so i don't see any downside

scripts/cangen/cangen/templates/bus.h.jinja2
```diff
@@ -0,0 +1,67 @@
...
+    {% set msg_title = "Rx"+msg.name %}
+
+    // Get the latest message of type {{msg_title}}
+    auto Get{{msg_title}}() -> std::optional<{{msg_title}}> {
+        etl::lock_guard<etl::mutex> lock(msg_guard_);
+        return {{ msg.name | camel_to_snake }}_msg_;
+    }
+
+    // Get the latest message of type {{msg_title}} and clear it
+    auto Pop{{msg_title}}() -> std::optional<{{msg_title}}> {
```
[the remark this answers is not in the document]
BlakeFreer: sure, I think of it more like a bucket that you can either look at or look at and clear. You can use `GetMessage` for the old behaviour but I have already found cases where I want to clear the old message

scripts/cangen/cangen/templates/messages.h.jinja2
```diff
@@ -100,82 +107,71 @@ private:
...
+    {% for sig in msg.signals %}
+    {% set sig_var = sig.name | camel_to_snake %}
+    {% set sig_var_type = signal_types[msg.name][sig.name] %}
+    static constexpr double k{{ sig.name }}Scale = {{ sig.scale }};
+    static constexpr double k{{ sig.name }}Offset = {{ sig.offset }};
+    {% endfor %}
 };
-{% endfor %}
 
-// clang-format on
```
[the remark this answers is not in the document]
BlakeFreer: shouldn't have remove it, adding this line bacl

scripts/cangen/cangen/templates/messages.h.jinja2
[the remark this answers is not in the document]
samparent97: I actually meant to remove the comment when I saw where it was used, disregard

scripts/cangen/cangen/templates/bus.h.jinja2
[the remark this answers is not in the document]
samparent97: Ah I see, the naming makes sense then

firmware/projects/Demo/CAN/Send/main.cc
samparent97: Sure then leave it out but still why not have the send method be templates and not require the user to pack the message

firmware/projects/Demo/CAN/Receive/platforms/cli/bindings.cc
samparent97: Ah that makes more sense

firmware/projects/Demo/CAN/Send/main.cc
BlakeFreer: Re-added concepts. All other convos are resolved so I'm going to merge

TylerStAmour (approved):
Wrong account lol

Issue #317, opened 2024-11-12
TMS Fan Speed control


BlakeFreer:
The TMS controls a fan to cool down the accumulator. Currently it turns on if the temp is above 0 degC which is always the case, but this means we are blowing warmer air into on accumulator. This heats up the accumulator rather than cooling it.

## Task
Change the fan control curve to only turn on when the accumulator is above atmospheric temperature. We don't have a sensor for ambient air temp so just assume 20°

PR #326, opened 2024-11-23
TMS Fan Speed control

Tegh25:
* Updated TMS fan to activate at 20 degC or higher

Closing #317. Set new points in the fan lut so that power is at 30 when temp is 20 and power is 0 for temp 19.

Can increase its accuracy by adding a decimal point shown below, originally no decimal point so I did not add it in this PR.
```cpp
const float fan_lut_data[][2] = {
    {19.9, 0}, // 19.9 instead of 19
    {20, 30},
    {50, 100},
};
```


BlakeFreer (approved):
thanks

PR #328, opened 2024-11-24
Added CAN Equation

ManushPatell:
@BlakeFreer . I've included the CAN bus equation. I'm unsure about how the input will look, so I have included the test case structure I used, but I imagine it will be different.


scripts/cangen/cangen/traffic_calculation.py
```diff
@@ -0,0 +1,61 @@
+class BusLoadCalculator:
```
BlakeFreer, line 1: Put a header comment at the top of the file with your name and date. Also add a link to your can bus article. see the other cangen files for a header example

scripts/cangen/cangen/traffic_calculation.py
```diff
@@ -0,0 +1,61 @@
...
+ 
+    def __init__(self, can_speed = int):
+        """"
```
BlakeFreer, line 4: Extra quote mark

scripts/cangen/cangen/traffic_calculation.py
```diff
@@ -0,0 +1,61 @@
+class BusLoadCalculator:
+ 
+    def __init__(self, can_speed = int):
```
BlakeFreer, line 3: This should be : int not = int

scripts/cangen/cangen/traffic_calculation.py
```diff
@@ -0,0 +1,61 @@
...
+        """"
+        Initialize bus load calculator with the CAN Speed (bits per second)
+        :param can_speed: The speed of the CAN Bus in bits per second
+        
+        """
+        self.can_speed = can_speed 
+    
+    def frame_length(self, frame_data: tuple) -> float:
```
BlakeFreer, line 11: This function name suggests that it calculates the length of a single frame, but you actually multiply it by the frequency. Either change the name of the function or multiple by the frequency in the other function (I think the second is the better option)

BlakeFreer (changes requested):

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,70 @@
...
+    Return Bus Load percentage
+    """
+
+    return (total_bits/ (can_speed))*100 #bus load formula
+
+if __name__ == "__main__":   #TEST CASE USED
+
+    total_bits = 0
+
+    can_speed = 500000 #set to 500kbaud
```
samparent97, line 53: Nothing stopping us from doing 1Meg baud. Maybe we put this info in the config.yaml?

Thoughts @BlakeFreer?

BlakeFreer: Speed is configured in ioc. Somehow this function will need to know that, maybe we use config.yaml, but it needs to be clear that config can't change the ioc

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+
+def calculate_bus_load(total_bits, can_speed):
+    """
+    Calculate bus load by dividing total bits by can_speed (default: 500,000bits)
+    Return Bus Load percentage
+    """
+
+    return (total_bits/ (can_speed))*100 #bus load formula
+
+def return_baud(messages, can_speed):
```
BlakeFreer, line 42: This function should be called "calculate_bus_load" since it is the primary function for this module.

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+
+    fixed_length = 44  #fixed length of CAN frame
+
+    variable_length = data_bytes * 9.6 #accounted for bit stuffing
+
+    total_frame_length = variable_length + fixed_length 
+
+    return total_frame_length
+
+def calculate_bits(message_length, frequency):
```
BlakeFreer, line 23: This doesn't need to be its own function. Just do the multiplication inline

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+    Calculate total bits by multiplying with frequency of message
+    :param message_length is float representing the message length
+    :param frequency is an int representing how often a message is transmitted per second
+    """
+
+    total_bits = message_length * frequency
+
+    return (total_bits)
+
+def calculate_bus_load(total_bits, can_speed):
```
BlakeFreer, line 34: This also doesn't need to be its own function, move the logic

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+def calculate_bits(message_length, frequency):
+    """
+    Calculate total bits by multiplying with frequency of message
+    :param message_length is float representing the message length
+    :param frequency is an int representing how often a message is transmitted per second
+    """
+
+    total_bits = message_length * frequency
+
+    return (total_bits)
```
BlakeFreer, line 32: Return isn't a function so para paranthesis aren't aren't needed

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+
+def calculate_bus_load(total_bits, can_speed):
+    """
+    Calculate bus load by dividing total bits by can_speed (default: 500,000bits)
+    Return Bus Load percentage
+    """
+
+    return (total_bits/ (can_speed))*100 #bus load formula
```
BlakeFreer, line 40: Remove these parentheses

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,66 @@
...
+
+def return_baud(messages, can_speed):
+
+    total_bits = 0
+    # Initialize the calculator
```
BlakeFreer, line 45: This comment is outdated and can be removed

BlakeFreer (changes requested):
Structure is looking better.

There are a lot of extra blank lines. Use blank lines to separate logical sections of code, not necessarily every line. For example there shouldn't be a blank line after a `for` loop starts, and the data length function doesn't need any. 

There will be a few more changes after these comments to set up proper testing and integration with the rest of cangen. You can start looking at poetry tests 

Reach out to @AndrewI26. He set up  automated code formatting with "ruff" which will help you

BlakeFreer:
I created a `cangen/tests/` folder and added instructions to the README. Add some more tests for your calculator. Try to break it. Maybe run a test that ensures `ZeroDivisionError` is returned if canspeed is zero. See the Pytest docs <https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions>

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@dataclass
+class CANMessage:
+    data_length: int
+    frequency: int
+
+
+def data_length(data_bytes):
+    fixed_length = 44  # fixed length of CAN frame
```
BlakeFreer, line 17: don't need this comment

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,42 @@
...
+    variable_length = data_bytes * 9.6  # accounted for bit stuffing
```
BlakeFreer, line 18: this is a good comment since it is not immediately clear by we `*9.6` instead of `*8`

actually I just realized that bit stuffing should apply to the whole message, not just the data portion. therefore we should have `*8` here and `*1.2` on the total length

BlakeFreer (changes requested):

docs/docs/firmware/can-traffic/index.md
```diff
@@ -38,22 +38,24 @@ Each CAN frame consists of several fields with fixed bit lengths, plus a variabl
 
 The maximum total frame length is:
 
-\[ 1 + 11 + 1 + 6 + 16 + 2 + 7 + 9.6x = 44 + 9.6x \text{ bits}\]
+\[ 1 + 11 + 1 + 6 + 16 + 2 + 7 + 9.6x = 44 + 8x \text{ bits}\]
```
BlakeFreer, line 41: fix the other 9.6x

docs/docs/firmware/can-traffic/index.md
```diff
@@ -38,22 +38,24 @@ Each CAN frame consists of several fields with fixed bit lengths, plus a variabl
...
 
 where \(x\) represents the number of bytes in the variable data field.
 
 --------------------------------------
 
 ## Total CAN Traffic
 
+We must multiply total bits per second by 1.2 to account for bit stuffing
```
BlakeFreer, line 49: good

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+
+@dataclass
+class CANMessage:
+    data_length: int
+    frequency: int
+
+    @property
+    def bit_count(self):
+        fixed_length = 44  # fixed length of CAN frame
+        variable_length = self.data_length * 8.0  # accounted for bit stuffing
```
BlakeFreer, line 18: this comment is incorrect since this calculation does not include stuffing. Notice how really simple comment can easily become outdated.

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class CANMessage:
+    data_length: int
+    frequency: int
+
+    @property
+    def bit_count(self):
```
BlakeFreer, line 16: you can simplify this function to 

`return 44 + 8 * self.data_length`

The extra variables don't really help with readability. It would help if you included the entire list of CAN components (like RTR, CONTROL, CRC, etc). I'm not suggesting you do that, but it is already clear from the equation which part is fixed and which is variable

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+        fixed_length = 44  # fixed length of CAN frame
```
BlakeFreer, line 17: don't need this comment

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+        return total_frame_length
+
+
+def calculate_bus_load(messages: list[CANMessage], can_speed: int) -> float:
+    """Calculates the bus load percentage based on CAN messages and bus speed
+
+    Parameters
+    ----------
+    messages : list[CANMessage]
+        A list of CANMessage objects, each containing `bit_count` and `frequency` attributes
```
BlakeFreer, line 29: This is not a helpful comment. You can remove this line. The previous line already describes that this is "a list of CANMessage objects." Also, you do not want to describe the internal attributes of that object. The purpose of an object is to encapsulate data, so this function should not need to know how the class works.

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+    can_speed : int
+        The CAN bus speed in bits per second (bps).
+
+    Returns
+    -------
+    float
+        Returns bus load as a percentage
+    """
+    if can_speed < 0:
```
BlakeFreer, line 38: This is bad practice. We will likely use your function to terminate the build process if the traffic is too high, but this line will incorrectly say that there is 0 bus load if the speed is `<0`

we should raise an error if less than __or equal__ to 0, not just if less than

Instead, you should raise an error. This will terminate the build process because an invalid value was provided.

The correct error type is `ValueError`. You should replace this block with

```python
if can_speed <= 0:
    raise ValueError(f"Invalid CAN speed {can_speed}. Must be positive.")
```

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+        print(f"Invalid CAN speed: {can_speed}")
+        return 0
+
+    total_bits = sum(message.bit_count * message.frequency for message in messages)
+
+    if total_bits < 0:
```
BlakeFreer, line 44: You can assume that the total_bits is non-negative. This is true assuming that bit_count and frequency are non-negative. This function is not the correct place to verify that, so let's defer those checks to somewhere else. Remove this block

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+        return 0
+
+    bus_load = ((total_bits) * 1.2) / can_speed * 100
+    return round(bus_load, 2)
```
BlakeFreer, line 48: Don't round here. Let the caller of the function decide how to format the data.

scripts/cangen/cangen/traffic.py
```diff
@@ -0,0 +1,48 @@
...
+    if can_speed < 0:
+        print(f"Invalid CAN speed: {can_speed}")
+        return 0
+
+    total_bits = sum(message.bit_count * message.frequency for message in messages)
+
+    if total_bits < 0:
+        return 0
+
+    bus_load = ((total_bits) * 1.2) / can_speed * 100
```
BlakeFreer, line 47: the 1.2 should probably go in the `total_bits` calculation to match your website article. after moving it, add a comment reminding what the 1.2 is for

Also remove the parenthesis - bedmas can handle this

scripts/cangen/tests/test_traffic.py
BlakeFreer, line 1: A single pytest "test" is represented by a function. You should separate your tests into separate functions.

The function name MUST start with `test` or else pytest will not run it. You can use your comments to guide the function name

Ex

```python
def test_bus_overload():
    assert calculate_bus_load(
        [
            CANMessage(data_length=64, frequency=100),
            CANMessage(data_length=32, frequency=200),
        ],
        100000,
    ) == pytest.approx(138.72, rel=1e-5)
```

scripts/cangen/tests/test_traffic.py
```diff
@@ -0,0 +1,65 @@
...
+    assert calculate_bus_load(
+        [
+            CANMessage(data_length=8, frequency=100),
+            CANMessage(data_length=5, frequency=50),
+        ],
+        500000,
+    ) == pytest.approx(3.60, rel=1e-2)
+
+    # zero frequency or negative frequency
+    assert calculate_bus_load(
```
BlakeFreer, line 15: After adding the `raise ValueError` check mentioned earlier, you should verify that this function raises an error.

See <https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions>

scripts/cangen/tests/test_traffic.py
```diff
@@ -0,0 +1,65 @@
...
+    ) == pytest.approx(0, rel=1e-2)
+
+    # bus overload
+    assert calculate_bus_load(
+        [
+            CANMessage(data_length=64, frequency=100),
+            CANMessage(data_length=32, frequency=200),
+        ],
+        100000,
+    ) == pytest.approx(138.72, rel=1e-5)
```
BlakeFreer, line 30: good use of `approx`

scripts/cangen/tests/test_traffic.py
```diff
@@ -0,0 +1,65 @@
...
+    try:
+        assert calculate_bus_load(
+            [
+                CANMessage(data_length=8, frequency=1000),
+                CANMessage(data_length=4, frequency=800),
+            ],
+            0,
+        ) == pytest.approx(0, rel=1e-5)
+    except ZeroDivisionError:
+        print("Cannot divide by zero baud")
```
BlakeFreer, line 51: printing doesn't do anything in pytest. You should verify that this raises an error. See earlier comment

BlakeFreer (changes requested):

BlakeFreer:
@ManushPatell Is this ready for re-review? You clicked resolve on all my comments but some of the changes weren't implemented

BlakeFreer (changes requested):
Almost ready to merge. Just want to format the code to be consistent.

I added instructions for formatting code to the `cangen/README`. Run that command and it should make some formatting changes to your files but will not affect the logic. Commit those changes and we're ready to merge.

FYI any sizable coding project will have formatting guidelines. For example, the firmware directory in `racecar/` uses `clang-format` and we have a GitHub Action which prevents code from being merged if it is unformatted. We could set up an Action for our python projects as well, but I don't think that's a good use of time, so we just have to be diligent and format the code manually with that command.

BlakeFreer (approved):
Still needs to be integrated into the cangen pipeline, but that will require a new CAN database format.

PR #333, opened 2024-11-26
CubeMX with CMake

### LINKED ISSUE #140: Update to use STM32CubeMX CMake generation
Our build system currently uses the Makefile generated by CubeMX when building a project for the STM32 platform. Ex, go into `racecar/firmware` and run

```bash
make PROJECT=Demo/Blink PLATFORM=stm32f677
```

Eventually, CMake reaches the `firmware/projects/Demo/Blink/platforms/stm32f767/cubemx/CMakeLists.txt` which includes `firmware/cmake/build_cubemx.cmake`.

`build_cubem.cmake` first calls `firmware/cmake/generate_cubemx.mk` to regenerate code from CubeMX if the `board_config.ioc` file is newer than the `Makefile` generated by CubeMX. We then append some custom targets to that makefile which allows us to extract variables rfom it into CMake, all so that we can compile the CubeMX source files in CMake. It is a big mess.

STM32CubeMX can now generate for CMake directly which will cut out most of this mess.

# Task

You will prototype on the `Demo/Blink` project.

1. Convert the project to generate for CMake instead of Makefile. To change this option, open the `.ioc` file in CubeMX and go to the Project Management tab.
2. Generate the code. Document what the CMake project looks like.
3. Link the CMake target to main in `platforms/stm32f767/CMakeLists.txt`. You are replacing what is currently the `driver` library target. You are free to edit the `platforms/stm32f767/cubemx/CMakeLists.txt` and `firmware/cmake/build_cubemx,cmake` files if necessary.
4. Show that the system builds properly.
5. Document the CMake integration in Markdown to add to <https://macformula.github.io/racecar/firmware/>

# Notes

We currently use a Makefile `generate_cubemx.mk` rather than CMake to generate code from CubeMX. This is because, with the current setup, the generation must occur at CMake "configure" time such that the sources are known before starting CMake's "build" stage. Using a separate Makefile is what allows us to only regenerate when `board_config` is updated. If you can integrate all of this in CMake, then great, but only if you ensure that CubeMX only generates when necessary (as this is a very slow process).

BlakeFreer:
closes #140 

# Key changes

## CubeMX generates for CMake instead of Makefile!
- This greatly simplifies the build system since we don't need to rip data out of the Makefile.
- Also improves robustness since we can use the generated flags instead of fixed flags in Toolchain.cmake
- On that note, `toolchain.cmake` is no longer needed as it is generated on-the-fly by cubemx. No other platforms used it so I completely removed it
- CubeMX generation is executed in `mcal_conf.cmake` (instead of `platforms/stm32cubemx/CMakeLists`) because we need the toolchain to exist before CMake starts looking at subfolders (specifically, before `project()` is called in `firmware/CMakeLists.txt`)

I had to do a few hacky things to the generated code because CubeMX is quite rigid. I modify cubemx's generated toolchain file (`gcc-arm-none-eabi.cmake`) to create `racecar-toolchain.cmake`:
- `CMAKE_*_COMPILER_FORCE` is commented out as it prevents clangd from finding system headers
- The linker path is modified since our `CMAKE_SOURCE_DIR` is not the cubemx dir (it is firmware)

These modifications happen automatically whenever cubemx regenerates so there is no impact or responsibility to the developer.

## Removes the `bindings` CMake target.

```cmake
#### BEFORE ####
# firmware/CMakeLists.txt
add_executable(main)
add_library(bindings)
...
target_link_libraries(main PRIVATE bindings)

#platforms/.../CMakeLists.txt
target_sources(bindings PRIVATE bindings.cc)
target_link_libraries(bindings PUBLIC mcal-stm32f767)
```
```cmake
#### AFTER ####
# firmware/CMakeLists.txt
add_executable(main)

#platforms/.../CMakeLists.txt
target_sources(main PRIVATE bindings.cc)
target_link_libraries(main PRIVATE mcal-stm32f767)
```
It was causing the [static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof) and causing duplicate compilation since sources which should have been compiled with main were compiled on the side.

Now, in platforms/.../CMakeLists.txt, all sources, mcal layers, and cubemx sources are linked directly to the `main` target

Also removes the `os` target - files are just linked to `main` directly. This removes this ugly block from `firmware/CMakelists.txt`

```cmake

# Only link/include RTOS if the project supports RTOS
# The project needs to provide an underlying OS target
# in the platform/${PLATFORM_NAME}/CMakeLists.txt
if(TARGET os)
    target_link_libraries(main PRIVATE os)
endif()
```

# Why make these changes?

The Makefile based system was very hacky. It only existed because cubemx did not support CMake generation until very recently.

## Why now?

Our projects are getting more complex and have more source files. The current system doesn't (easily) allow for you to break mcal or shared code into `.h` and `.cc`. Since mcal and shared were independent CMake targets, they did not inherit the compiler flags required for `stm` so they could not be linked to the other files.

In particular, this is necessary for my work on the CAN drivers which, for circular dependency reasons, require the CAN sources be split into header and source files.

# Known Issues

Clangd is getting confused by some of the file names. In particular, CubeMX generates peripheral files like `gpio.h`, while we have `shared/periph/gpio.h` and `mcal/stm32f767/periph/gpio.h`. Part of clangd's strategy is to associate source and header files with similar names (ie. the same prefix, as far a I know), so clangd thinks that our c++ `gpio.h` files are actually C files. **THIS DOES NOT AFFECT COMPILATION**, only the language server. I looked through the clangd docs and forums and couldn't find a solution. There's no clangd flag or anything which disables this behaviour.

To fix this, we will need to rename our header files. This means either changes the name ex `gpio.h -> shared_gpio.h` or the extension `gpio.h -> gpio.hpp`. I'm in favour of the latter since it is less intrusive, and it guarantees that we won't ever collide with a C file (in case cubemx decides to add `shared_gpio.c`). EDIT: See #334 

# Testing

@samparent97 @TylerStAmour @CameronBeneteau @Corbin-Jarrett @sokosam @AmanRV @luaibash @Tegh25 

This commit includes the relevant updates to all of the projects and their platforms, with the exception of raspi/sil platforms. I have tested the `stm32f767` platforms on the board and they work, but I want some people to try the new build system on their windows / mac machines and verify it all compiles.

You must run deepclean before building to clear out the old Cubemx files

`make PROJECT=... PLATFORM=... deepclean`

Replace the ... as needed

Please add a PR comment with your OS, which projects/platforms you tested, whether it worked, and any other notes


sokosam:
OS: Windows 11
Tested: Everything in Demo except RTOS
Platform: CLI
Compiled and works as intended

TylerStAmour (changes requested):
The debug projects are missing these changes (even the non-sil ones), so they won't compile

BlakeFreer:
> The debug projects are missing these changes (even the non-sil ones), so they won't compile

Yes I didn't convert them. Those projects will be removed in the next few weeks so I'm not worried about them.

CameronBeneteau (approved):
OS: macOS Sequoia (v15.1.1)

Successfully built all Demo projects for cli and stm32 (no need for GNU sed).

Tegh25:
OS: Windows 10
Builds were successful for all Demo, FrontController, LVController, and TMS projects for the cli and stm platforms.

Corbin-Jarrett (commented):
Integration of CubeMX CMake looks good.

Testing: Windows 11
Platform: stm32
Projects: Demo/Blink, Demo/RTOS, Demo/BasicIO
All projects built with no errors.

BlakeFreer (commented):
need to update a few comments

luaibash:
OS: Windows 11
Tested: All demos, and FrontController, LVController, and TMS projects
Platform: CLI
Everything compiled as expected.

PR #334, opened 2024-11-26
Rename headers to `.hpp`

BlakeFreer:
rename all of our `.h` files to `.hpp` to address the clangd the issue mentioned in #333 

Updates documentation website to mimic this change.

I built all projects and platforms (excluding SIL and the debug projects) so I don't need anyone to check those.

DO NOT MERGE until `freerb/mx_cmake` (PR #333) is merged. I don't want to make that PR any harder to read


BlakeFreer (commented):
Just renaming files and updating docs to match

PR #343, opened 2024-11-30
LV for EV6

BlakeFreer:
Update LV controller. Removing some old abstractions like `Indicator` and `StateBroadcaster` to simplify the code.

~~Uses a functional approach to process Contactors messages~~ Changed this, see [this comment](https://github.com/macformula/racecar/pull/343#issuecomment-2610024657)


BlakeFreer:
I went back on my functional approach to processing the std::optional contactor state messages.

This was the "functional" approach. It defines `logic` to process the message (`.transform(logic)`) and returns false if the data isn't there.

```c++
bool IsContactorsOpen() {
    auto logic = [](generated::can::RxContactorStates contactor_states) {
        return (contactor_states.PackPositive() == 0 &&
                contactor_states.PackNegative() == 0 &&
                contactor_states.PackPrecharge() == 0);
    };

    // If the contactor states are not available, return false.
    return veh_can.GetRxContactorStates().transform(logic).value_or(false);
}
```

While that is cool, it uses some syntax that isn't very common, like lambdas and understanding optionals as Monads. I settled on this simplier implementation

```c++
bool IsContactorsClosed() {
    auto contactor_states = veh_can.GetRxContactorStates();

    if (contactor_states.has_value()) {
        return (contactor_states->PackPositive() == 1 &&
                contactor_states->PackNegative() == 1 &&
                contactor_states->PackPrecharge() == 0);
    } else {
        return false;
    }
}
```

BlakeFreer (commented):
lgtm

PR #352, opened 2024-12-22
Implements bitset function from ETL

tamaraxu:


BlakeFreer (changes requested):
Good improvements and it's good to see more tests with complicated DBC messages.

I'm adding some details which I didn't mention clearly enough in the issue. The purpose of using a `bitset` is it already "knows" how to access values at arbitrary positions in itself.

The idea for Rx Unpack() is
1. Convert the RawMessage.data byte array to an etl::bitset
2. Use the extract method to directly access ranges in the bitset. use the compile-time `extract` method documented at <https://www.etlcpp.com/bitset.html>
3. Scale / offset the valeus (if necessary)

Based on the bitset documentation, there doesn't seem to be function which can replace the Tx `Pack` method bit operations, so let's not modify that.

scripts/cangen/cangen/templates/messages.hpp.jinja2
```diff
@@ -76,18 +77,35 @@ private:
...
         {% set sig_var_scale = "k" + sig.name + "Scale" %}
         {% set sig_var_offset = "k" + sig.name + "Offset" %}
+        {% set counter = 0 %}
         {% for mask in masks %}
             {% set i = loop.index0 %}
             {% set shift = shifts[i] %}
             {% if mask != 0 %}
                 {% if shift >= 0 %}
-        {{ temp_sig_var }} |= shared::can::unpack_left_shift<{{ temp_sig_var_type }}>(raw_msg.data[{{ i }}], {{ shift }}U, {{ mask | decimal_to_hex }}U);
+        etl::bitset<8 + {{ shift }}U> temp_bitset_{{ sig_var | camel_to_snake }}_{{ i }}(raw_msg.data[{{ i }}]);
```
BlakeFreer, line 86: The point of bitset is that the **entire** message is **one** bitset. Not a bunch of individual little bitsets.

For an 8 byte message, you should convert the whole message to a 64 bit bitset then extract out of it.

scripts/cangen/cangen/templates/messages.hpp.jinja2
```diff
@@ -76,18 +77,35 @@ private:
...
+        constexpr etl::bitset<8 + {{ shift }}U> temp_mask_{{ sig_var | camel_to_snake }}_{{ i }}({{ mask | decimal_to_hex }}U);
+        constexpr size_t count_mask_{{ sig_var | camel_to_snake }}_{{ i }} = temp_mask_{{ sig_var | camel_to_snake }}_{{ i }}.count();
```
BlakeFreer, line 88: This line seems needlessly complicated. If it is constexpr, why don't you just set the value? Seems like you're using bitset.count just to get back to a known integer value

BlakeFreer (changes requested):

BlakeFreer:
This PR shouldn't include my CAN driver commits. Please fix the tree locally then force push.

BlakeFreer:
Closing this PR since there hasn't been any development in several months and the unpacking logic has since been improved by #451

PR #368, opened 2024-12-30
Format string

ManushPatell:
- Implemented std::format in C++ to the appropriate files. I wasn't able to check all the files through testing but the demo for BasicIO and Blink worked as intended
- *Created the branch off of a previous branch so the previous branches commits are there* <-- did rebase thanks to Blake


firmware/mcal/cli/periph/can.h
```diff
@@ -26,18 +27,13 @@ class CanBase : public shared::periph::CanBase {
     }
 
     void Send(const shared::can::RawCanMsg& can_tx_msg) {
-        std::cout << iface_ << " [" << std::hex << std::uppercase
-                  << std::setfill('0') << std::setw(2) << can_tx_msg.header.id
-                  << "] ";
+        "[ "<< std::format("{} {:02X}", iface_ can_tx.msg.header.id) << "] ";
```
BlakeFreer, line 30: Where does this get printed? Why aren't the square brackets in the format string? Also I don't think the format string is equivalent to the old statement

firmware/mcal/cli/periph/can.h
```diff
@@ -26,18 +27,13 @@ class CanBase : public shared::periph::CanBase {
...
 
         // Loop through each data byte and print it in uppercase hex with
         // leading zeros
         for (int i = 0; i < sizeof(can_tx_msg.data); ++i) {
-            std::cout << std::hex << std::uppercase << std::setfill('0')
-                      << std::setw(2) << static_cast<int>(can_tx_msg.data[i])
-                      << " ";
+            std::cout << std::format("{:02X} ", static_cast<int>(can_tx_msg.data[i]));
```
BlakeFreer, line 35: Do we need static cast still?

firmware/mcal/cli/periph/gpio.h
```diff
@@ -19,10 +20,10 @@ class DigitalInput : public shared::periph::DigitalInput {
 
     bool Read() override {
         int value;
-        std::cout << "Reading DigitalInput " << name_ << std::endl;
+        std::cout <<, std::format("Reading DigitalInput {}", name_ ) << std::endl;
```
BlakeFreer, line 23: What's the comma for? Looks like it'll cause an error

firmware/mcal/cli/periph/gpio.h
```diff
@@ -19,10 +20,10 @@ class DigitalInput : public shared::periph::DigitalInput {
...
         std::cout << " | Enter 0 for False, 1 for True: ";
         std::cin >> value;
-        std::cout << " | Value was " << (value ? "true" : "false") << std::endl;
+        std::cout << std::format("| Value was {}", (value ? "true": "false" )) << std::endl;
```
BlakeFreer, line 26: Remove () around the ternary statement. Also lost a space in the string

firmware/mcal/cli/periph/gpio.h
```diff
@@ -35,8 +36,7 @@ class DigitalOutput : public shared::periph::DigitalOutput {
     DigitalOutput(std::string name) : name_(name) {}
 
     void Set(bool value) override {
-        std::cout << "Setting DigitalOutput Channel " << name_ << " to "
-                  << (value ? "true" : "false") << std::endl;
+        std::cout << std::format("Setting DigitalOutput Channel {} to {}", name_, (value ? "true": "false") ) << std::endl;
```
BlakeFreer, line 39: Remove () around ternary

BlakeFreer, line 39: Check the cpp format string spec. Do we need the ternary operator or is there a simpler way to print a string version of a Boolean

firmware/mcal/cli/periph/pwm.h
```diff
@@ -16,37 +17,33 @@ class PWMOutput : public shared::periph::PWMOutput {
     PWMOutput(std::string name) : name_(name) {}
 
     void Start() override {
-        std::cout << "Starting PWM " << name_ << std::endl;
+        std::cout << std::format("Starting PWN {}", name_) << std:endl; 
     }
 
     void Stop() override {
-        std::cout << "Stopping PWM " << name_ << std::endl;
+        std::cout << std::format("Stopping PWN {}", name_) << std::endl;
```
BlakeFreer, line 24: Spelling

firmware/mcal/cli/periph/pwm.h
```diff
@@ -16,37 +17,33 @@ class PWMOutput : public shared::periph::PWMOutput {
...
+        std::cout << std::format("Stopping PWN {}", name_) << std::endl; 
     }
 
     void SetDutyCycle(float duty_cycle) override {
         duty_cycle_ =
             shared::util::Clamper<float>::Evaluate(duty_cycle, 0, 100);
 
-        std::cout << "Setting PWM " << name_ << " duty cycle to " << duty_cycle_ << "%"
-                  << std::endl;
+        std::cout << std::format("Setting PWN {} duty cycle to {:.3f} %", name_, duty_cycle_) << std::endl;
```
BlakeFreer, line 31: Spelling

BlakeFreer, line 31: No space after percent. Good use of .3f but that'll print a lot of decimals like 93.123. Is there a way to print 3 sig figs instead? (Don't just change to .1f)

BlakeFreer (changes requested):

firmware/mcal/cli/periph/can.h
ManushPatell: i think so, std::format doesn't seem to do anything to support the conversion to an int. I think without the cast, it would interpret it as an unsigned char and therefore give the ascii

BlakeFreer: ok, have you tested this? what happens if you remove static cast?

firmware/mcal/cli/periph/gpio.h
```diff
@@ -19,10 +20,10 @@ class DigitalInput : public shared::periph::DigitalInput {
 
     bool Read() override {
         int value;
-        std::cout << "Reading DigitalInput " << name_ << std::endl;
+        std::cout << std::format("Reading DigitalInput {}", name_) << std::endl;
         std::cout << " | Enter 0 for False, 1 for True: ";
         std::cin >> value;
-        std::cout << " | Value was " << (value ? "true" : "false") << std::endl;
+        std::cout << std::format(" | Value was {}", value ? "true": "false") << std::endl;
```
BlakeFreer, line 26: can the boolean expression be simplified?

BlakeFreer (changes requested):
one small note, otherwise good.

firmware/mcal/cli/periph/can.h
ManushPatell: I didn't notice anything different when I removed the static cast

firmware/mcal/cli/periph/gpio.h
BlakeFreer: Unresolving, this should be simplified. Look at what you've done for `DigitalOutput::Set()`

firmware/mcal/cli/periph/can.h
BlakeFreer: Ok then let's remove it. No point having useless code

BlakeFreer (changes requested):

BlakeFreer:
Still need to update:
- mcal/cli/periph: adc.h, analog_output.h

firmware/mcal/cli/periph/analog_output.h
```diff
@@ -15,8 +15,7 @@ class AnalogOutput : public shared::periph::AnalogOutput {
     AnalogOutput(std::string name) : name_(name) {}
 
     void SetVoltage (float voltage) override {
-        std::cout << "Setting AnalogOutput " << name_ << " to " << voltage << " V" 
-                  << std::endl;
+        std::cout << std::format("Setting AnalogOutput {} to {} V", name_, voltage) << std::endl;
```
BlakeFreer, line 18: floating point numbers should almost always be rounded for printing. use 3 decimal places for voltage.

BlakeFreer (changes requested):

PR #369, opened 2025-01-04
Convert AMK Block from Simulink to C++

luaibash:
## Description
- This PR is to move the AMK motor block from simulink to C++.


BlakeFreer (commented):
This code is going to run on FrontController so we don't need another project. Can you put your 3 `amk_` files to `projects/FrontController/inc`? See what Teghveer has done in the `user/atelieyt/vd-interface` branch

BlakeFreer:
The AMK Update function is empty. Should there be anything there yet?

luaibash:
> The AMK Update function is empty. Should there be anything there yet?

@BlakeFreer I haven't put anything as I was simply just copying the structure that you suggested from the `tarun/battery_monitor` branch. I assume that's the first step to do once this is an okay structure to proceed with and there are no mistakes.

BlakeFreer:
> > The AMK Update function is empty. Should there be anything there yet?
> 
> @BlakeFreer I haven't put anything as I was simply just copying the structure that you suggested from the `tarun/battery_monitor` branch. I assume that's the first step to do once this is an okay structure to proceed with and there are no mistakes.

Sorry yes, that was the first step. Looks good. CAN config is set up, though you don't need the `veh` bus for this project since all of these messages are on powertrain `pt.dbc`. Also, notice how the left and right AMK's are handled almost, if not, identically in Simulink. I highly suggest you abstract out that common logic and just run it on the left/right inputs.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+            output_status = MiStatus::ERROR;
+        }
+        else if (left_motor_status == right_motor_status) {
+            output_status = left_motor_status;
+        }
+        else {
+            output_status = previous_state_status;
+        }
+
+        previous_state_status = output_status;
```
BlakeFreer, line 244: you don't actually need separate `previous_state_status` and `output_status` variables. You can just modify the instance variable and remove the else branch since this will leave the state as is.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+                update_motor_output.inverter = 0;
+
+                set_point.amk__target_velocity = 0;
+                set_point.amk__torque_limit_positiv = 0;
+                set_point.amk__torque_limit_negativ = 0;
+                // Rest of setpoint values do not get set
+                // I'm thinking we have to have private var hold previous output, set SP set_point = previous_set_point at start of UpdateMotor
+                // This way on every call of UpdateMotor, if an output value is not set, it's cause the prev is retained
+
+                // Set some time condition with after(500, msec)
```
BlakeFreer, line 206: this is where you should use `time_ms`. store a class variable for the time when you enter this state and perform the transition if `time-time_entered > 500`

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+                set_point.amk__target_velocity = motor_input.speed_request;
+                set_point.amk__torque_limit_positiv = motor_input.torque_limit_positive;
+                set_point.amk__torque_limit_negativ = motor_input.torque_limit_negative;
+
+                if (val1.amk_b_error == 1) {
+                    amk_state = AmkStates::ERROR_DETECTED;
+                }
+                else if (cmd == MiCmd::SHUTDOWN) {
+                    amk_state == AmkStates::SHUTDOWN;
```
BlakeFreer, line 191: accidental double `==`

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+                    amk_state = AmkStates::READY;
+                }
+
+                break;
+
+            case AmkStates::READY:
+                update_motor_output.status = MiStatus::READY;
+                update_motor_output.inverter = 0;
+                
+                if (val1.amk_b_error == 1) {
```
BlakeFreer, line 171: In C++, use `if (val)` or `if(!val)` for boolean types, not `== 1` or `== 0` like Simulink.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+                update_motor_output.status = MiStatus::ERROR;
+
+                if (cmd == MiCmd::ERR_RESET) {
+                    amk_state = AmkStates::ERROR_RESET;
+                }
+
+                break;
+
+            case AmkStates::ERROR_RESET:
+                // Create new function for nested state machine block
```
BlakeFreer, line 220: since the nested state machine is so simple (it's linear), just flatten it and make one big state machine

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+                break;
+
+            case AmkStates::SHUTDOWN:
+                update_motor_output.inverter = 0;
+
+                set_point.amk__target_velocity = 0;
+                set_point.amk__torque_limit_positiv = 0;
+                set_point.amk__torque_limit_negativ = 0;
+                // Rest of setpoint values do not get set
+                // I'm thinking we have to have private var hold previous output, set SP set_point = previous_set_point at start of UpdateMotor
```
BlakeFreer, line 203: yes good idea

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+                set_point.amk__torque_limit_negativ = 0;
+
+                if (cmd == MiCmd::STARTUP) {
+                    amk_state = AmkStates::STARTUP;
+                }
+
+                break;
+
+            case AmkStates::STARTUP:
+                // Create new function for nested state machine block
```
BlakeFreer, line 156: the internal state machine is simple enough. just flatten it into one big state machine

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+struct MotorInput {
+    double speed_request;
+    double torque_limit_negative;
+    double torque_limit_positive;
+};
+
+// AmkInput struct, the input that is fed into the state machine
+struct AmkInput {
+    MiCmd cmd;
+    generated::can::AMK0_ActualValues1 amk_actual_values_1_left;
```
BlakeFreer, line 37: you can choose to shorten these var names, may like 

```
amk_actual_values_1_left -> left_actual1
amk_setpoints_1_right -> right_setpoints // there's setpoints 2 so we don't need the _1_
```

There is enough context for the reader to figure out more details if they need. The longer variable names make it harder to follow through the logic

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+private:
+    AmkStates amk_state;
+    MiStatus previous_state_status = MiStatus::OFF;
+
+    template<AmkActualValues1 V1, AmkActualValues2 V2, SetPoint SP>
+    UpdateMotorOutput<SP> UpdateMotor(const V1 val1, const V2 val2, const MotorInput motor_input, const MiCmd cmd) {
+        UpdateMotorOutput<SP> update_motor_output;
+        SP set_point;
+
+        switch (amk_state) {
```
BlakeFreer, line 136: With a single state machine, there is a temporal mismatch between the current state and the output, since the output is set _before_ transitioning the state.

For example, after the first transition, state=STARTUP but status is still OFF.

To fix this, I recommend using 2 steps - one to handle transitions and the other to handle outputs.

For DriverInterface, I created a `Transition()` function [here](https://github.com/macformula/racecar/blob/manush/driver-interface/firmware/shared/controls/driver_interface_fsm.hpp#L69) which, given the necessary inputs, determines the next state. Then, you can use that state to determine the outputs.

Also notice how you can handle multiple equivalent transitions with this structure since you're not limited to a basic switch-case. You can handle all 6 `AMK_b_error == 1` transitions together which simplifies the code (see the first lines of that Transition() function)

BlakeFreer (changes requested):

BlakeFreer:
make sure to get clang-format working too. lmk if you have questions about it

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -1,50 +1,119 @@
...
+    generated::can::AMK1_SetPoints1 amk_setpoints_1_right;
     bool inverter;
 };
 
-// Only DECLARE the class methods an variables here. Write the method
-// definitions in the .cc file.
+// AmkActualValues1 concept, combines AMK0_ActualValues1 and AMK1_ActualValues1 into one common type for easier use
+template <typename MsgType>
+concept AmkActualValues1 = requires(MsgType msg) {
+    { msg.amk_b_system_ready }; // TODO: Add the std::same_as<bool> check, currently it is erroring out
```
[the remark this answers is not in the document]
BlakeFreer: fixed

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -1,50 +1,119 @@
...
+
+// Make enum of all possible states in the simulink FSM and then add cases logic and transitions
+
 class AmkBlock {
 public:
-    // do not change the public interface
     AmkBlock();
-    AmkOutput update(const AmkInput& input, int time_ms);
+    AmkOutput update(const AmkInput& input, int time_ms) {
+        UpdateMotorOutput<generated::can::AMK0_SetPoints1> left_update_motor_output = UpdateMotor<generated::can::AMK0_ActualValues1, generated::can::AMK0_ActualValues2, generated::can::AMK0_SetPoints1>(input.amk_actual_values_1_left, input.amk_actual_values_2_left, input.left_motor_input, input.cmd);
```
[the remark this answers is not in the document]
luaibash: Modified both variable types to use auto.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+    ERROR_RESET
+};
+
+class AmkBlock {
+public:
+    AmkBlock(AmkStates initial_state) : amk_state(initial_state) {}
+
+    AmkOutput update(const AmkInput& input, int time_ms) {
+        UpdateMotorOutput<generated::can::AMK0_SetPoints1> left_update_motor_output = UpdateMotor<generated::can::AMK0_ActualValues1, generated::can::AMK0_ActualValues2, generated::can::AMK0_SetPoints1>(input.amk_actual_values_1_left, input.amk_actual_values_2_left, input.left_motor_input, input.cmd);
+        UpdateMotorOutput<generated::can::AMK1_SetPoints1> right_update_motor_output = UpdateMotor<generated::can::AMK1_ActualValues1, generated::can::AMK1_ActualValues2, generated::can::AMK1_SetPoints1>(input.amk_actual_values_1_right, input.amk_actual_values_2_right, input.right_motor_input, input.cmd);
```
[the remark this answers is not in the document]
luaibash: Modified variables names to just be `left` and `right`.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Yes, the new decision is to flatten the state machine, add new states under AmkStates, and add them under the switch statement.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Renamed these variables to be shorter.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+    ERROR_DETECTED,
+    ERROR_RESET
+};
+
+class AmkBlock {
+public:
+    AmkBlock(AmkStates initial_state) : amk_state(initial_state) {}
+
+    AmkOutput update(const AmkInput& input, int time_ms) {
+        UpdateMotorOutput<generated::can::AMK0_SetPoints1> left_update_motor_output = UpdateMotor<generated::can::AMK0_ActualValues1, generated::can::AMK0_ActualValues2, generated::can::AMK0_SetPoints1>(input.amk_actual_values_1_left, input.amk_actual_values_2_left, input.left_motor_input, input.cmd);
```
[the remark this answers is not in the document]
luaibash: Added namespace declaration.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Good catch

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Switched all if statements that use bools to not explicitely check value.

firmware/projects/AmkBlock/inc/amk_block.cc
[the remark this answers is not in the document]
luaibash: I currently moved all functions to the .cpp file, why do the templated functions need to stay in the header file?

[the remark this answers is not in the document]
BlakeFreer: the compiler needs to know how to implement the functions during preprocessing. I assume for laying out the function memory.

You won't notice the error because your CMake isn't set up correctly. don't worry about the cmake for now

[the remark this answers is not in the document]
luaibash: Sounds good, moved the template function back to the header file.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Created a Transition method and moved all transition logic to it.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Added `elapsed_time` calculations in `Transition` method.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,247 @@
...
+        UpdateMotorOutput<generated::can::AMK1_SetPoints1> right_update_motor_output = UpdateMotor<generated::can::AMK1_ActualValues1, generated::can::AMK1_ActualValues2, generated::can::AMK1_SetPoints1>(input.amk_actual_values_1_right, input.amk_actual_values_2_right, input.right_motor_input, input.cmd);
+
+        AmkOutput output {
```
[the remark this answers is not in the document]
luaibash: I would say this no longer applies since I am now setting the output to the `previous_state_output_` and then returning it, so it cannot be immediately returned.

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: With the new implementation of `previous_state_output_`, I use this method to now be called when `previous_state_output_` is set:

```
previous_state_output_ = {
        .status = ProcessOutputStatus(left.status, right.status),
        .left_setpoints = left.setpoints,
        .right_setpoints = right.setpoints,
        .inverter_enable = left.inverter_enable && right.inverter_enable};
```

So I'm going to keep the return of this method to be more readable.

firmware/projects/AmkBlock/inc/amk_block.cc
```diff
@@ -0,0 +1,56 @@
...
+
+    UpdateMotorOutput<generated::can::AMK1_SetPoints1> right = {
+        .setpoints = previous_state_output_.right_setpoints,
+        .status = previous_state_output_.status,
+        .inverter_enable = previous_state_output_.inverter_enable,
+        .amk_state = right_amk_state_,
+        .amk_state_start_time = right_amk_state_start_time_};
+    UpdateMotor<AMK1_ActualValues1, AMK1_ActualValues2, AMK1_SetPoints1>(
+        input.right_actual1, input.right_actual2, input.right_motor_input,
+        input.cmd, time_ms, right);
```
luaibash, line 28: I'm allowed to remove the <...> specifications for each of these statements, let me know if you would like this or keep it as is.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,381 @@
...
+
+// UpdateMotorOutput struct, used to create a return type for UpdateMotor such
+// that it does not matter if setpoints is for left or right
+template <SetPoints SP>
+struct UpdateMotorOutput {
+    SP setpoints;
+    MiStatus status;
+    bool inverter_enable;
+    AmkStates amk_state;
+    int amk_state_start_time;
```
luaibash, line 119: Added these two variables so this struct can be passed into `UpdateMotor` and have every single value be updated to return as an output and to update the current state/start time of either the left or right motor.

The main reason I did this was because `amk_state_` and `amk_state_start_time_` had to be split into two each, one for left and one for right. `UpdateMotor` had no way of knowing which to update, so I had to make the function return the value to be set in `Update`.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,381 @@
...
+private:
+    AmkStates left_amk_state_;
+    AmkStates right_amk_state_;
+    int left_amk_state_start_time_ = 0;
+    int right_amk_state_start_time_ = 0;
+    AmkOutput previous_state_output_{
+        .status = MiStatus::OFF,
+        .left_setpoints = generated::can::AMK0_SetPoints1{},
+        .right_setpoints = generated::can::AMK1_SetPoints1{},
+        .inverter_enable = 0};
```
luaibash, line 139: Split `amk_state_` and `amk_state_start_time_` into left and right values as they are separate.

Also let me know if you'd rather have default values defined here or in the init constructor.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,381 @@
...
+
+    template <AmkActualValues1 V1, AmkActualValues2 V2, SetPoints SP>
+    void UpdateMotor(const V1 val1, const V2 val2, const MotorInput motor_input,
+                     const MiCmd cmd, int time_ms,
+                     UpdateMotorOutput<SP>& update_motor_output);
+    template <AmkActualValues1 V1, AmkActualValues2 V2, SetPoints SP>
+    void Transition(const V1 val1, const V2 val2, const MotorInput motor_input,
+                    const MiCmd cmd, const int time_ms,
+                    UpdateMotorOutput<SP>& update_motor_output);
```
luaibash, line 148: Modified both methods to be void return type and to take in `update_motor_output`.

Again, the main reason I did this was because amk_state_ and amk_state_start_time_ had to be split into two each, one for left and one for right. UpdateMotor had no way of knowing which to update, so I had to make the function return the value to be set in Update.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,381 @@
...
+    MiStatus ProcessOutputStatus(MiStatus left_status, MiStatus right_status);
+};
+
+// Computes main logic for updating the motor output given motor inputs
+template <AmkActualValues1 V1, AmkActualValues2 V2, SetPoints SP>
+void AmkBlock::UpdateMotor(const V1 val1, const V2 val2,
+                           const MotorInput motor_input, const MiCmd cmd,
+                           const int time_ms,
+                           UpdateMotorOutput<SP>& update_motor_output) {
```
luaibash, line 157: It seems like V2/val2 is never used inside the FSM, do we even need it?

firmware/projects/AmkBlock/inc/amk_block.hpp
BlakeFreer: we have two nearly identical functions operating with nearly identical instance variables.

The clean solution is to make a separate class for MotorUpdate, then hold two instances under your AmkBlock. Then you can have the separate left and right inputs/outputs AND instance variables. 

You should structure this subclass identically to the AmkBlock.

This also makes testing easier since you can just walk through the state flow of one motor at a time, so your AmkBlock testing will only have the combined logic

firmware/projects/AmkBlock/inc/amk_block.hpp
BlakeFreer: Nope. Take it out then.

firmware/projects/AmkBlock/inc/amk_block.hpp
BlakeFreer: See comment about making a separate class.

firmware/projects/AmkBlock/inc/amk_block.hpp
BlakeFreer: See comment about a separate class

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Can we remove `generated::can::AMK0_ActualValues2 left_actual2` and generated::can::AMK1_ActualValues2 right_actual2; out of `AmkInput` as well as the concept for them, or does the input still have to atleast have those to follow the simulink model?

BlakeFreer: No we can remove that as well. We will keep it in the DBC for now

firmware/projects/AmkBlock/inc/amk_block.hpp
luaibash: Created two total classes `MotorInterface` and `AmkManager`.

firmware/projects/AmkBlock/inc/amk_block.cc
```diff
@@ -0,0 +1,29 @@
+#include "amk_block.hpp"
+
+MotorInterface::MotorInterface() {}
```
luaibash, line 3: This init constructor has nothing so lmk if you'd rather still have it explicitely defined or not.

firmware/projects/AmkBlock/inc/amk_block_test.cc
```diff
@@ -0,0 +1,294 @@
...
+    assert(actual_sp.amk__target_velocity == expected_sp.amk__target_velocity);
+    assert(actual_sp.amk__torque_limit_positiv ==
+           expected_sp.amk__torque_limit_positiv);
+    assert(actual_sp.amk__torque_limit_negativ ==
+           expected_sp.amk__torque_limit_negativ);
+}
+
+// Tests the whole control model sequence from start to end
+void test_normal_sequence() {
+    UpdateMotorOutput<generated::can::AMK0_SetPoints1> output;
```
luaibash, line 21: I made all tests use AMK0, which is the left configurations. I don't like having to specify whether to use the left or right type but I think we have to here.

Let me know if the structure of the tests are okay and if we need any other tests.

BlakeFreer (changes requested):

firmware/projects/AmkBlock/inc/amk_block_test.cc
BlakeFreer: this is fine. templated classes don't exist on their own - they're "templates" to create concrete instances given a template parameter. there's no way around this

the nice thing is we use the same function `UpdateMotorOutput` for both sides, so the right side is guaranteed to work because of the concept. The only way it could fail is if the AMK1 classes had a different structure that the AMK0, but the concept would detect this and prevent compilation, so it can't fail

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,377 @@
...
+public:
+    MotorInterface();
+
+    AmkOutput Update(const AmkInput& input, const int time_ms);
+
+private:
+    AmkManager<generated::can::AMK0_SetPoints1> left_amk_manager{
+        AmkStates::MOTOR_OFF_WAITING_FOR_GOV};
+    AmkManager<generated::can::AMK1_SetPoints1> right_amk_manager{
+        AmkStates::MOTOR_OFF_WAITING_FOR_GOV};
```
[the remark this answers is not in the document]
luaibash: I don't think setting namespaces inside a class is possible, I get errors thrown and searching it up online says I cannot.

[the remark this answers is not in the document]
luaibash: Since there are two parameters for the template now, it makes these initializations even more ugly. We can set the whole file to possibly be set under that namespace to avoid it if you'd like.

firmware/projects/AmkBlock/inc/amk_block.hpp
```diff
@@ -0,0 +1,377 @@
...
+            previous_state_output_.status = MiStatus::OFF;
+            previous_state_output_.inverter_enable = 0;
+
+            previous_state_output_.setpoints.amk_b_inverter_on = 0;
+            previous_state_output_.setpoints.amk_b_dc_on = 0;
+            previous_state_output_.setpoints.amk_b_enable = 0;
+            previous_state_output_.setpoints.amk_b_error_reset = 0;
+            previous_state_output_.setpoints.amk__target_velocity = 0;
+            previous_state_output_.setpoints.amk__torque_limit_positiv = 0;
+            previous_state_output_.setpoints.amk__torque_limit_negativ = 0;
```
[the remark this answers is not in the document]
luaibash: I don't think this is possible since the two possible setpoint types are classes, and I can't just straight up define their values like a struct. I tried and it's throwing errors.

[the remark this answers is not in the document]
BlakeFreer: shoot ok. can you at least do this?

```
output = {
    .status = ....
    .inverter = ...
    .setpoints.amk_b = ...
    .setpoints.amk_b = ...
    .setpoints.amk_b = ...
};
```

firmware/projects/AmkBlock/inc/amk_block.hpp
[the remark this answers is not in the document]
BlakeFreer: yes you're right. it's possible to rename individual types:

```
using AMK0_ActualValues1 = generated::can::AMK0_ActualValues1;
```

but that's not worth it for this. keep as is

firmware/projects/AmkBlock/inc/amk_block.hpp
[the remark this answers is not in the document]
luaibash: No that doesn't work either.

firmware/projects/AmkBlock/inc/amk_block.cc
```diff
@@ -0,0 +1,28 @@
...
+
+    return AmkOutput{
+        .status = ProcessOutputStatus(left.status, right.status),
+        .left_setpoints = left.setpoints,
+        .right_setpoints = right.setpoints,
+        .inverter_enable = left.inverter_enable && right.inverter_enable};
+}
+
+// Processes the new output status given the left/right statuses
+MiStatus MotorInterface::ProcessOutputStatus(MiStatus left_status,
```
BlakeFreer, line 18: I realized this function has a bit of a strange flow. It assumes two responsibilities: updating the `status_` field and returning a value for the `AmkOutput`.

It's weird to return an instance variable since the object can just access it.

I think a clearer solution is to rename the method `UpdateOutputStatus`, have it return nothing, and just modify the `status_ ` field internally. 

Then we would call `UpdateStatus` before constructing the AmkOutput and instead set `.status = status_`

firmware/projects/AmkBlock/inc/amk_block.hpp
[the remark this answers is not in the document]
BlakeFreer: ok, we'll keep as is

firmware/projects/AmkBlock/inc/amk_block.cc
luaibash: Hmm okay. I defined it this way since I thought it would be more readable to have it return the output status so that when returning the ouput in the update function, the reader could see that the way we calculate the status value is through this method:

```
return AmkOutput{
        .status = ProcessOutputStatus(left.status, right.status),
        .left_setpoints = left.setpoints,
        .right_setpoints = right.setpoints,
        .inverter_enable = left.inverter_enable && right.inverter_enable};
```

If you think this doesn't add any clarity, I will go ahead and implement the suggestion you made, it's a simple change.

BlakeFreer: this is fine if you think that it's clear that the method modifies the object's internal state.

luaibash: Eh actually I went back on what I said, it's better practice to separate the responsibilites and have this method only update the instance. I've implemented your suggestion.

BlakeFreer (approved):
Great Work!

PR #371, opened 2025-01-05
User/mfang/silpolymorphism

fangm16:
Created the private functions in sil_client.cc and sil_client.h:

## RegisterIO (std::string ecu_name, std::string sig_name, signals::SignalDirection sig_dir, signals::SignalType sig_type)
Handles common logic between the RegisterDigitalInput, RegisterDigitalOutput, and RegisterAnalogInput functions. 
There are differences in how each Register function uses Signal Direction and Signal Type, therefore pass it in as an argument. 

## ReadRegister(std::string ecu_name, std::string sig_name, signals::SignalDirection sig_dir, signals::SignalType sig_type, signals::ReadSignalResponse read_response);

Handles common logic between ReadDigitalLevel and ReadAdcVoltage
There are multiple differences between these two functions, so I decided to leave some code unabstracted. 
The initialization of read_response is left unabstracted, therefore needing a pass by reference. 

## SetRegister(std::string ecu_name, std::string sig_name, bool level);
Only one function requires the SetRegister function, but still implemented an abstracted version for future use. 

---

Update 01-08:
I am figuring out how to compile with gRPC and protobuf

### DIFF
--- a/firmware/validation/sil/sil_client.cc
+++ b/firmware/validation/sil/sil_client.cc
@@ -167,7 +167,7 @@ bool SilClient::ReadDigitalLevel(std::string ecu_name, std::string sig_name) {
 
     ReadRegister(ecu_name, sig_name, 
         signals::SignalDirection::SIGNAL_DIRECTION_INPUT, 
-        signals::SignalType::SIGNAL_TYPE_ANALOG, &read_response);
+        signals::SignalType::SIGNAL_TYPE_ANALOG, read_response);
 
     return read_response.value_digital().level();
 
@@ -202,7 +202,7 @@ double SilClient::ReadAdcVoltage(std::string ecu_name, std::string sig_name) {
 
     ReadRegister(ecu_name, sig_name, 
         signals::SignalDirection::SIGNAL_DIRECTION_INPUT, 
-        signals::SignalType::SIGNAL_TYPE_ANALOG, &read_response);
+        signals::SignalType::SIGNAL_TYPE_ANALOG, read_response);
 
     if (!read_response.has_value_analog()) {
         std::cout << "expected adc value got: " << read_response.value_case()
--- a/firmware/validation/sil/sil_client.cc
+++ b/firmware/validation/sil/sil_client.cc
@@ -7,6 +7,7 @@
 
 #include <iostream>
 #include <string>
+#include <format>
 
 #include "signals.grpc.pb.h"
 #include "signals.pb.h"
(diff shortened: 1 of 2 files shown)


BlakeFreer:
Clean up git. Includes several irrelevant commits

firmware/validation/sil/sil_client.cc
BlakeFreer, line 1: whats with all the commented out lines? They should be deleted if we don't need them anymore

firmware/validation/sil/sil_client.cc
```diff
@@ -23,68 +23,205 @@ void SilClient::Connect() {
...
+        std::cout << "expected adc value got: " << read_response.value_case()
                   << std::endl;
     }
+
+    return read_response.value_analog().voltage();
 }
 
-void SilClient::RegisterAnalogInput(std::string ecu_name,
-                                    std::string sig_name) {
+void RegisterIO(std::string ecu_name, std::string sig_name, signals::SignalDirection sig_dir, signals::SignalType sig_type) {
```
BlakeFreer, line 215: RegisterIO should be a member function. Same with Set/ReadRegister

firmware/validation/sil/sil_client.cc
```diff
@@ -124,65 +260,31 @@ void SilClient::SetDigitalLevel(std::string ecu_name, std::string sig_name,
...
     read_request.set_signal_name(sig_name);
-    read_request.set_signal_type(signals::SignalType::SIGNAL_TYPE_ANALOG);
-    read_request.set_signal_direction(
-        signals::SignalDirection::SIGNAL_DIRECTION_INPUT);
+    read_request.set_signal_type(sig_type);
+    read_request.set_signal_direction(sig_dir);
 
-    status = stub_->ReadSignal(&context, read_request, &read_response);
+    status = stub_->ReadSignal(&context, read_request, read_response);
     if (!status.ok()) {
```
BlakeFreer, line 276: update all `<<` string building to use `std::format` instead.

firmware/validation/sil/sil_client.cc
```diff
@@ -124,65 +260,31 @@ void SilClient::SetDigitalLevel(std::string ecu_name, std::string sig_name,
...
         std::cout << status.error_code() << ": " << status.error_message()
                   << std::endl;
         return 0.0;
     }
 
     if (!read_response.status()) {
-        std::cout << "write response error: " << read_response.error()
+        std::cout << "write response error: " << read_response->error()
                   << std::endl;
         return 0.0;
```
BlakeFreer, line 285: shouldn't return a value from a void function

firmware/validation/sil/sil_client.cc
```diff
@@ -124,65 +260,31 @@ void SilClient::SetDigitalLevel(std::string ecu_name, std::string sig_name,
...
+    status = stub_->ReadSignal(&context, read_request, read_response);
     if (!status.ok()) {
         std::cout << status.error_code() << ": " << status.error_message()
                   << std::endl;
         return 0.0;
     }
 
     if (!read_response.status()) {
-        std::cout << "write response error: " << read_response.error()
+        std::cout << "write response error: " << read_response->error()
```
BlakeFreer, line 283: why did this change from `.` to `->`?

BlakeFreer (changes requested):
Left a few comments. There are several little syntactic errors that would be detected by the compiler. Can you figure out how to compile with grpc and protobuf?

Here's some details for grpc. https://github.com/grpc/grpc/blob/v1.66.0/src/cpp/README.md

firmware/validation/sil/sil_client.cc
fangm16: im not sure to be honest, I think I was messing around and forgot to change it back

BlakeFreer:
Looks good. Have you got it to compile? That's the next step then we can merge

fangm16:
> Looks good. Have you got it to compile? That's the next step then we can merge

Not yet, there might be some issues with my CMake configuration but I'm also still figuring out how to compile with gRPC

BlakeFreer:
closed as stale

Issue #355, opened 2024-12-26
DBC to JSON


BlakeFreer:
### Discussed in https://github.com/macformula/racecar/discussions/275

<div type='discussions-op-text'>

<sup>Originally posted by **BlakeFreer** November  1, 2024</sup>
What might this look like? How and where will it be parsed? Are there existing standard formats?</div>

BlakeFreer:
UBC uses JSON for their CAN database. Again, this issue is not about JSON vs YAML, but you can see how they chose to structure their database. I'm not endorsing it in any way, but it may be a source of inspiration

BlakeFreer:
Check out the `racecar/datasheets/amk` for a description of the AMK motor CAN messages. Our database obviously needs to support those motor messages.

luaibash:
Here are my initial thoughts with reading this issue:

- `@dataclass` definitely does help reduce boilerplate code, so let's go ahead with using that for defining the classes.
- UBC's [CAN database implementation](https://github.com/UBCFormulaElectric/Consolidated-Firmware/tree/master/scripts/code_generation/jsoncan) is pretty useful for understanding what we want to do and what it could look like.

Responses to questions:

1. **Should we have two separate databases:** I think we should keep these separate, as it follows better coding practice.
2. **Which parameters should be optional:** I don't see any parameters that should be optional yet, but what I'm thinking with this is that we set optional parameters that have default values if the value is not provided by the instance.
3. **The message receivers are specified per signal. Does this make sense:** I think it's a good idea to try and "set the receiver(s) for the entire message to simplify the database". Taking a quick look at the messages in [pt.dbc](https://github.com/macformula/racecar/blob/main/firmware/projects/pt.dbc), it looks like that would be the simpler implementation as all signals within the same message use the same receiver. We just have to ask ourselves, could there ever be a scenario where a message could have multiple receivers? If no, is it something we want to support for the future?

BlakeFreer:
> could there ever be a scenario where a message could have multiple receivers?

This is very possible. The TMS temperature message is used by the BMS and FrontController. But is there ever a message for which some (say 5 of 8) signals are used by a node and other (3 of 8) signals aren't? I don't think so, so let's specify a list of receiver nodes per message, not per signal as dbc does

Jinwoo-H:
Here are some of my thoughts

I agree we should keep the two databases separate

[...]

PR #375, opened 2025-01-10
CAN database structure draft

Jinwoo-H:
Here's some design decisions I took
- `scale` and `offset` are optional, as they are not used for `INTEGRAL` signals
    - Should we have a post-initialization hook to validate that these values are not defined for `INTEGRAL` signals?
- `unit` is optional, as many signals have empty strings for units (e.g. see `veh.dbc`)
- Enums are stored at the Database level via a dictionary

### DIFF
--- a/scripts/cangen/cangen/database.py
+++ b/scripts/cangen/cangen/database.py
@@ -1,42 +1,25 @@
 from __future__ import annotations
 from dataclasses import dataclass, field
 from enum import Enum, auto
-from typing import Dict, List, Set, Optional, Type, Union
-
-class SignalType(Enum):
-    INTEGRAL = auto()
-    FLOATING = auto()
+from typing import Dict, List, Set, Optional, Type
 
 class Endian(Enum):
     BIG = auto()
     LITTLE = auto()
 
-@dataclass
-class Signal:
-    name: str # 1-32 chars, only [A-z], digits, and underscores
-    start_bit: int # bit 0-63
-    length: int # 1-64 bits
-    endianness: Endian
-    is_signed: bool
-    minimum: Union[int, float]
-    maximum: Union[int, float]
-    unit: Optional[str] = None
-    signal_type: SignalType
-    receivers: Set[Node]
-    description: Optional[str] = None
-
-    # Not needed for INTEGRAL signals
-    scale: Optional[float] = None
-    offset: Optional[float] = None
+class IntegralType(Enum):
+    BOOLEAN = auto()
+    INTEGER = auto()
+    UNSIGNED_INTEGER = auto()
+    ENUM = auto()
 
 @dataclass
-class Message:
-    name: str # 1-32 chars, only [A-z], digits, and underscores
-    id: int # 0-2047 (11-bits) or 0-536870911 (29-bits for extended IDs)
-    length: int # 0-8 bytes
-    sender: Node
-    signals: List[Signal]
-    frequency: int
+class Bus:
+    name: str
+    baud_rate: float
+    nodes: Set[Node]
+    messages: List[Message]
+    enum_types: Dict[str, Type[Enum]] = field(default_factory=dict) 
     description: Optional[str] = None
 
 @dataclass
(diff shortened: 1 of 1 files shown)


scripts/cangen/cangen/database.py
BlakeFreer, line 1: nit: `from __future__ import annotations` allows you to reorder the classes. Can you reorganize to follow the hierarchy? i.e. `Bus` at the top and `Signal` near the bottom.

This will make it easier for newcomers to understand the structure as they read from top-down

scripts/cangen/cangen/database.py
```diff
@@ -0,0 +1,59 @@
...
+    name: str # 1-32 chars, only [A-z], digits, and underscores
+    start_bit: int # bit 0-63
+    length: int # 1-64 bits
+    endianness: Endian
+    is_signed: bool
+    minimum: Union[int, float]
+    maximum: Union[int, float]
+    unit: Optional[str] = None
+    signal_type: SignalType
+    receivers: Set[Node]
```
BlakeFreer, line 25: Regarding [your comment](https://github.com/macformula/racecar/issues/355#issuecomment-2575286894), I agree with maintaining the granularity of receivers-per-signal, but this will become tedious since nearly all signals of a message have the same receiver(s).

Can we move `receivers` to `Message` and add an optional `additional_receivers` to the `Signal`? This gives us both efficiency and granularity. We will merge the two sets per-signal to create the DBC

scripts/cangen/cangen/database.py
```diff
@@ -0,0 +1,59 @@
...
+
+@dataclass
+class Node:
+    name: str
+    description: Optional[str] = None
+
+@dataclass
+class Bus:
+    name: str
+    baud_rate: int
```
BlakeFreer, line 50: nit: while baud usually is a whole number, I think it makes more sense as a `float` since it will be used in division calculations

scripts/cangen/cangen/database.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
+from dataclasses import dataclass, field
+from enum import Enum, auto
+from typing import Dict, List, Set, Optional, Type, Union
+
+class SignalType(Enum):
```
BlakeFreer, line 6: I think `SignalType` has the potential to grow a lot in the future, such as for error messages and alerts, so we should expand this beyond a simple enum to incorporate type-specific fields.

- A floating signal should include the scale / offset, unit, min / max, is_signed.
- An integral signal should specify whether it is a boolean, integer, unsigned integer, or which enum it uses (then we can verify that the signal bit-length can accomadate all enum values).
- An error signal may specify its own "private" enum that doesn't need to be shared

You don't need to implement the error signal, but my point it that expanding `SignalType` to a set of related dataclasses gives us the flexibility later on.

Also, it solves the problem of the Optional fields in the Signal. min/max, unit, scale/offset are all features of a floating-point, measured signal and should be restricted as such.

You may choose to use either inheritance `class FloatingSignal(Signal)` or composition `class Signal: sig_type: FloatingSignal`

BlakeFreer (changes requested):

scripts/cangen/cangen/database.py
BlakeFreer: This addresses your comment

> Should we have a post-initialization hook to validate that these values are not defined for INTEGRAL signals?

scripts/cangen/cangen/database.py
```diff
@@ -0,0 +1,59 @@
...
+    name: str # 1-32 chars, only [A-z], digits, and underscores
+    id: int # 0-2047 (11-bits) or 0-536870911 (29-bits for extended IDs)
+    length: int # 0-8 bytes
+    sender: Node
+    signals: List[Signal]
+    frequency: int
+    description: Optional[str] = None
+
+@dataclass
+class Node:
```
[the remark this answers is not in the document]
Jinwoo-H: I agree it would simplify things, but I'm thinking we keep it as we want a defined way to represent our nodes, similar to how in a .dbc file we define them like `BU_: AMK1 AMK0 FC`. Let me know your thoughts

scripts/cangen/cangen/database.py
Jinwoo-H: Thanks for the feedback, I've implemented `FloatingSignal` and `IntegralSignal` as child classes. Do we want the parent `Signal` class to be an abstract class then? In other words, do we want to enforce that signals are always implemented as a specified signal type?

scripts/cangen/cangen/database.py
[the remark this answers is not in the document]
BlakeFreer: Ok, let's keep the Node class

scripts/cangen/cangen/database.py
BlakeFreer: Yes let's make `Signal` abstract.

scripts/cangen/cangen/database.py
BlakeFreer, line 1: You've implemented the Signal->ConcreteSignal relationship well. The child class structure will give us a lot of space to expand functionality in the future.

Now that I see it, I think we should make separate classes for `BooleanSignal` and `EnumSignal` since they are sufficiently different from `IntegerSignal`. For example, it doesn't make sense to specify `value_range` for those two, and `EnumSignal` needs a config parameter for the name of the enumeration it will be using (i.e. the key in `bus.enum_types`).

Then we can reserver `IntegerSignal` for pure integer counting values.

BlakeFreer, line 1: This doesn't matter much, but why do some fields have `= something`?

BlakeFreer (changes requested):

BlakeFreer:
Andrew set up `ruff` to format our code. Assuming you are using vscode, you can install the `Ruff` extension from Astral, set it as the default formatter for python files in this workspace, and enable format on save to autoformat your code ([instructions](https://code.visualstudio.com/updates/v1_6#_editor)).

PR #357 will reject pull requests if the code is not properly formatted, so setting up autoformat will avoid this

scripts/cangen/cangen/database.py
Jinwoo-H: I have the optional fields `= None` so the parameter is optional when we instantiate an object (i.e. we don't have to explicitly pass `None`)

Similar thing with `enum_types`, however I use a `default_factory` which ensures that all instances of `Bus` gets its own dict, as opposed to sharing it if it was `= {}`

Everything else I don't provide a default value since I am assuming they will all be provided during creation

BlakeFreer: Good explanation, thanks

BlakeFreer (approved):

PR #380, opened 2025-01-12
Tarun/battery monitor

Nagatarun25:
I implemented the 2 state machines (battery_monitor and bm_control). I need to create another test sequence to test the bm_control outputs.


BlakeFreer (changes requested):
Excellent work

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,298 @@
...
+                break;
+            }
+            if (input.hvil_status == false) {
+                current_status = BmStatus::HVIL_INTERRUPT;
+                snapshot_time_ms = time_ms;
+                std::cout
+                    << "[DEBUG] Transitioning from RUNNING to HVIL_INTERRUPT\n";
+                break;
+            }
+            if (input.precharge_contactor_states != false ||
```
BlakeFreer, line 145: I talked to @dorlanj-0. In the running state, if the contactor states are not POS=true, NEG=true, PRE=false, then we should go back to the INIT state.

You should modify the Control update fsm to also go back to Startup if state == INIT

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,71 @@
...
+    CLOSE_HV_NEG,
+    CLOSE_PRECHARGE,
+    CLOSE_HV_POS,
+    OPEN_PRECHARGE,
+};
+
+enum ContactorSates {
+    OPEN = false,
+    CLOSED = true,
+};
```
luaibash, line 32: What is this for, is it being used or going to be used in a future commit?

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,244 @@
...
+    output.status = status;
+    output.contactor = contactor_cmd;
+
+    return output;
+}
+
+// Control update method
+BmStatus BatteryMonitor::bmStatusUpdate(const BmInput& input, int time_ms) {
+    BmOutput output;
+    output.contactor = {false, false, false};
```
luaibash, line 31: What is this for?

firmware/projects/FrontController/inc/battery_monitor.h
luaibash: Ah okay no worries you can take your time.

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,71 @@
...
+
+private:
+    // add private methods / variables as necessary.
+    BmStatus bmStatusUpdate(const BmInput& input, int time_ms);
+    ContactorCMD bmControlUpdate(BmStatus status, int time_ms);
+
+    // State machine variables (BmUpdate)
+    BmStatus current_status;
+    BmControlStatus bm_control_status;
+    int snapshot_time_ms;
```
luaibash, line 70: These should have an underscore at the end of each name: https://google.github.io/styleguide/cppguide.html#Variable_Names

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+
+    if (input.pack_soc < 30) {
+        current_status = BmStatus::LOW_SOC;
+        std::cout << "[DEBUG UPDATE] Transitioning from IDLE to LOW_SOC\n";
+    } else if (input.hvil_status == OPEN) {
+        current_status = BmStatus::HVIL_INTERRUPT;
+        snapshot_time_ms = time_ms;
+        std::cout
+            << "[DEBUG UPDATE] Transitioning from IDLE to HVIL_INTERRUPT\n";
+    }
```
luaibash, line 39: Looking at the simulink model, it looks like these transitions should only occur for every state except for the INIT state. But right now it seems like these transitions could occur for any state, including INIT.

Let me know if there is something by design that I am missing or if it is implemented incorrectly.

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+                current_status = BmStatus::INIT;
+
+                break;
+            }
+            break;
+
+        case BmStatus::HVIL_INTERRUPT:
+            if (time_ms - snapshot_time_ms >= 5000) {
+                // TODO: do we need to check every 5 seconds
+                if (input.hvil_status) {
```
luaibash, line 112: I think this should be using that `OPEN/CLOSED` notation.

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+        case BmStatus::IDLE:
+            if (input.precharge_contactor_states == OPEN &&
+                input.hv_neg_contactor_states == CLOSED &&
+                input.hv_pos_contactor_states == OPEN) {
+                current_status = BmStatus::STARTUP;
+                std::cout
+                    << "[DEBUG UPDATE] Transitioning from IDLE to STARTUP\n";
+                break;
+            }
+            break;
```
luaibash, line 61: You don't need two break statements for this and the other cases, only one outside the if statement is all you need.

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+                std::cout << "[DEBUG BM CONTROL UPDATE] Transitioning from "
+                             "STARTUP_CMD to CLOSE_HV_NEG\n";
+                break;
+            }
+            break;
+
+        case BmControlStatus::CLOSE_HV_NEG:
+
+            if (time_ms - snapshot_time_ms >= 100) {
+                if (current_status == BmStatus::STARTUP) {
```
luaibash, line 153: No point using nested if statements, may as well AND them

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+
+        case BmStatus::LOW_SOC:
+            std::cout << "[DEBUG UPDATE] In LOW_SOC state\n";
+            break;
+    }
+
+    return current_status;
+}
+
+ContactorCMD BatteryMonitor::bmControlUpdate(BmStatus status, int time_ms) {
```
luaibash, line 129: If you would like, this function can be separated into two different functions `bmControlTransition` and `bmControlUpdate`, where the transition will deal with the first half of the code, setting the `bm_control_status`, and the update will deal with the entry actions and setting up the output.

Here is an example of how [blake](https://github.com/macformula/racecar/blob/020f5d7c7b303d9592afd5b8a6b2dbee92447b89/firmware/shared/controls/driver_interface_fsm.hpp#L49) implemented it in his code. Your choice on whether you want to split it up or not, Blake and I have opted to split it up in our code.

firmware/projects/FrontController/inc/battery_monitor.cc
Nagatarun25: on my previous commit, Blake told me to move the INIT state into the startupsuperstate, so if HVIL=0 or SOC < 30, we go the the corresponding states, rather than wait until entering the startup state. He also told me to remove HVIL ==1 from the transition between INIT to IDLE since it must be true or else we would begin with interrupt state. 

Another comment he made was that I should handle the superstate transitions all at once before the switch-case statement starts. Let me know if I still need to fix anything regarding this.

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,235 @@
...
+        bm_control_status = BmControlStatus::STARTUP_CMD;
+        std::cout << "[DEBUG BM CONTROL UPDATE] Transitioning from "
+                     "CLOSE_HV_NEG to STARTUP_CMD\n";
+    }
+
+    switch (bm_control_status) {
+        case BmControlStatus::STARTUP_CMD:
+            if (current_status == BmStatus::IDLE) {
+                bm_control_status = BmControlStatus::CLOSE_HV_NEG;
+                snapshot_time_ms = time_ms;
```
luaibash, line 143: Instead of setting snapshot_time_ms everytime a new state is switched to, you can set a temporary variable such as `new_state` which you can initialize to the current `bm_control_status` and set the new status to in each of these case statements. And then after the case statements you just add a check of the following:

```
if (new_state != bm_control_status) {
 snapshot_time_ms = time_ms;
}
```

This way you are setting the snapshot only once instead of having extra duplicated code. Lmk if that makes sense or you need a better explanation.

firmware/projects/FrontController/inc/battery_monitor.cc
luaibash: Ahhh okay that makes sense then, yes this all looks correct then, the handling of all transitions at the start is good!

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+
+enum class BmControlStatus {
+    STARTUP_CMD,
+    CLOSE_HV_NEG,
+    CLOSE_PRECHARGE,
+    CLOSE_HV_POS,
+    OPEN_PRECHARGE,
+};
+
+enum ContactorSates : bool {
```
BlakeFreer, line 29: typo, should be `ContactorState`

also this should be an `enum class`. This will force you to write `ContactorState::OPEN` everywhere, but you can get around that with `using enum ContactorState;` at the top of each relevant method

you should also update the ContactorCMD and BmInput structs so the fields are `ContactorState` type instead of bool

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,214 @@
...
+        bm_control_status_ = new_state;
+        snapshot_time_ms_ = time_ms;
+    }
+
+    bmControlTransition(contactor_cmd);
+
+    return contactor_cmd;
+}
+
+void BatteryMonitor::bmControlTransition(ContactorCMD& cmd) {
```
BlakeFreer, line 192: this isn't a "transition," it sets the outputs. you should use a different name

also this method should not take in a command to modify. it completely sets the struct so use

`return ContactorCMD {OPEN, CLOSED, ...};` 

instead of

```
cmd = {...};
break;
....
return cmd;
```
your method should probably take `bm_control_status_` as an input. i know  we have access to it as a member field, but this will improve encapsulation

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,214 @@
...
+
+        case BmStatus::LOW_SOC:
+            std::cout << "[DEBUG UPDATE] In LOW_SOC state\n";
+            break;
+    }
+
+    return current_status_;
+}
+
+ContactorCMD BatteryMonitor::bmControlUpdate(BmStatus status, int time_ms) {
```
BlakeFreer, line 122: this method should be called `...Transition`. see comment below

BlakeFreer, line 122: you aren't using the `status` argument. should you be?

you are switching based on the `bm_control_status_` member field which never gets assigned to

this method should return the new state. then in the `.Update()` method, use that state to figure out the ContactorCMD. see comments on that function below

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,214 @@
...
+                current_status_ == BmStatus::PRECHARGE) {
+                new_state = BmControlStatus::OPEN_PRECHARGE;
+                std::cout << "[DEBUG BM CONTROL UPDATE] Transitioning from "
+                             "CLOSE_HV_POS to OPEN_PRECHARGE\n";
+            }
+
+            break;
+
+        case BmControlStatus::OPEN_PRECHARGE:
+            if (contactor_cmd.precharge_contactor_cmd == CLOSED ||
```
BlakeFreer, line 173: this should be based on the contactor status, not command. (notice that you always initialize the cmd to {open,open,open} above so this if would never pass

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,214 @@
...
+            }
+            break;
+    }
+
+    if (new_state != bm_control_status_) {
+        bm_control_status_ = new_state;
+        snapshot_time_ms_ = time_ms;
+    }
+
+    bmControlTransition(contactor_cmd);
```
BlakeFreer, line 187: as above, move this statement out into `.Update()` and return `new_state_` instead.

BlakeFreer (changes requested):

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,212 @@
+#include "battery_monitor.h"
+
+#include <iostream>
+
+// Constructor
```
BlakeFreer, line 5: remove these comments. they don't add any value (I can just read the method name)

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,212 @@
...
+BatteryMonitor::BatteryMonitor()
+    : current_status_(BmStatus::INIT),
+      snapshot_time_ms_(0),
+      bm_control_status_(BmControlStatus::STARTUP_CMD) {}
+
+// Update method
+BmOutput BatteryMonitor::update(const BmInput& input, int time_ms) {
+    BmOutput output;
+
+    // BM Status Update
```
BlakeFreer, line 15: remove these comments too

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+    BmCmd cmd;
+    ContactorState precharge_contactor_states;
+    ContactorState hv_pos_contactor_states;
+    ContactorState hv_neg_contactor_states;
+    ContactorState hvil_status;
+    float pack_soc;
+};
+
+struct ContactorCMD {
+    ContactorState precharge_contactor_cmd;
```
BlakeFreer, line 44: don't need to repeat the typename in the fields. rename to `precharge`, `hv_positive`, `hv_negative`

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+    ContactorCMD contactor;
+};
+
+// Only DECLARE the class methods and variables here. Write the method
+// definitions in the .cc file.
+class BatteryMonitor {
+public:
+    // do not change the public interface
+    BatteryMonitor();
+    BmOutput update(const BmInput& input, int time_ms);
```
BlakeFreer, line 60: rename to `Update`

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+
+private:
+    // add private methods / variables as necessary.
+    BmStatus bmStatusUpdate(const BmInput& input, int time_ms);
```
BlakeFreer, line 64: rename to `TransitionStatus`

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+    BmControlStatus bmControlTransition(BmStatus status, int time_ms);
```
BlakeFreer, line 65: rename to `TransitionControl`

firmware/projects/FrontController/inc/battery_monitor.h
```diff
@@ -0,0 +1,72 @@
...
+    ContactorCMD setContactorCMD(BmControlStatus bm_control_status_);
```
BlakeFreer, line 66: rename to `SelectContactorCmd`

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,212 @@
...
+            break;
+    }
+
+    return current_status_;
+}
+
+BmControlStatus BatteryMonitor::bmControlTransition(BmStatus status,
+                                                    int time_ms) {
+    using enum ContactorState;
+    // ContactorCMD contactor_cmd = {OPEN, OPEN, OPEN};
```
BlakeFreer, line 127: remove this line

firmware/projects/FrontController/inc/battery_monitor.cc
```diff
@@ -0,0 +1,212 @@
...
+            return ContactorCMD{CLOSED, OPEN, CLOSED};
+
+        case BmControlStatus::CLOSE_HV_POS:
+            return ContactorCMD{CLOSED, CLOSED, CLOSED};
+
+        case BmControlStatus::OPEN_PRECHARGE:
+            return ContactorCMD{OPEN, CLOSED, CLOSED};
+    }
+
+    return ContactorCMD{OPEN, OPEN, OPEN};
```
BlakeFreer, line 211: the switch-case covers all possible enums, so it is guaranteed to return. remove this

firmware/projects/FrontController/inc/battery_monitor_test.cc
```diff
@@ -0,0 +1,198 @@
...
+            .pack_soc = 20.0,  // SOC below LOW_SOC threshold
+        },
+        time_ms);
+
+    assert(output7.status == BmStatus::LOW_SOC);
+    assert(output7.contactor.precharge_contactor_cmd == OPEN);
+    assert(output7.contactor.hv_pos_contactor_cmd == OPEN);
+    assert(output7.contactor.hv_neg_contactor_cmd == OPEN);
+
+    // Test: HVIL_INTERRUPT and back to INIT
```
BlakeFreer, line 152: based on the simulink model, we should not be able to exit the LOW_SOC state, so the test should end here.

you should copy the sequence to have two separate tests:
1. init to low_soc
2. init to running to hvil to init

you will need to update the logic in the StatusUpdate block to prevent the system from leaving the LOW_SOC state via the `if(hvil_status == OPEN)` condition

BlakeFreer (changes requested):
Few things to reformat / rename based on our style guide. https://google.github.io/styleguide/cppguide.html#Function_Names

also one note about the state flow / test sequence since there's a difference from simulink

BlakeFreer (approved):

PR #388, opened 2025-01-17
Convert the Governor from Simulink

BlakeFreer:
# Checklist

- [X] Implement the Governor
- [X] test regular behaviour
- [x] add tests for the error conditions

# Building and testing

```bash
make PROJECT=Governor PLATFORM=cli
./build/Governor/cli/main
```


firmware/projects/Governor/governor.hpp
```diff
@@ -0,0 +1,33 @@
...
+        DiCmd di_cmd;
+        GovSts gov_sts;
+    };
+
+    Governor();
+    Output Update(const Input input, const int time_ms);
+
+private:
+    Output state_;
+    int state_entered_time_;
```
luaibash, line 26: I guess it works either way but might wanna explicitely initialize the state's time value to 0

BlakeFreer: i thought about this. yes, currently the class will implicitly initialize it, but I don't want to imply that 0 is the correct "starting value." the starting value isn't really defined until the first `.Update()` call when it is set to `time_ms`

I think the better solution is to add an `unititialized` state for the FSM to start in. Then we can define an unconditional transition from `uninit` to the existing `INIT` state on the first `Update()`
```c++
switch(state){
    case UNINIT:
        state = INIT; // no condition - this transition will always happen
        break;
```
Then we can definitively initialize the `entered_time` variable, without worrying about a default value being used by any non-`uninitialized` state

BlakeFreer: See my latest commit https://github.com/macformula/racecar/pull/388/commits/b1931ae97a812e66b585a3ba917ec30f4d79941c

Instead of adding a new enum value for uninitialized, I represent the state with `std::optional` so `nullopt` (i.e. empty) represents that the first transition has not yet occurred.

This makes an explicit transition to INIT during the first update call so the `state_entered_time` variable is guaranteed to have a valid value before it can be used

luaibash: Yes I like that change, removes the need for `is_first_update_` which is nice.

luaibash (approved):

PR #396, opened 2025-02-02
File parser for CAN DBC replacement

### LINKED ISSUE #356: DBC to JSON - File Parsing
# Goals

- Create a file representation of the database structure defined by #355.
- Convert all the `racecar/firmware` DBCs to this file format.
- Parse the files in Python.

- Write _good_ Python. Data parsing is a really interesting project since there are a lot of ways to use first-class functions and decorators to improve the behaviour without cluttering the code.
    - At any point, if you feel that you are repeating code, conditions, loops, error handling, etc, find a way to abstract out that behaviour. Our parsing code should focus more on the data structure than parsing.

## Non-goals

- Write the bare file parser. Instead, choose a file format that has existing Python support like json, yaml, toml, etc.

## Requirements

- The data should be validated as it is parsed. For example, we should ensure that all Signal bits fit inside of the message's data length without overlap.

- Any invalid / missing / unused value should cause an error during parsing.

# Resources

We currently use `cantools` to parse our DBCs. Look for inspiration in their database code. https://github.com/cantools/cantools/tree/master/src/cantools/database/can

@AndrewI26 wrote this parser for the `cangen` yaml configuration. https://github.com/macformula/racecar/blob/main/scripts/cangen/cangen/config.py
Notice how it first reads the yaml to a dictionary, then parses the dictionary for keys. This is good design since we could easily switch from yaml to json because both can be read to a dictionary.

Jinwoo-H:
Closes #356


firmware/projects/Demo/CAN/demo.json
BlakeFreer, line 1: This is an easy change later, but I think we should use yaml over json. I converted the pt & veh jsons to yaml and the files have ~30% less lines, plus i think yaml is easier to write since it doesn't have all of the `{ } [] " "` syntax

```json
{
  "name": "pt",
  "baud_rate": null,
  "nodes": [
    {
      "name": "AMK1"
    },
    {
      "name": "AMK0"
    },
    {
      "name": "FC"
    }
  ]
```

```yaml
name: pt
baud_rate: null
nodes:
  - name: AMK1
  - name: AMK0
  - name: FC
```

converted with `yq` 
```bash
choco install yq
yq -p json -o yaml pt.json >> pt.yaml
```

firmware/projects/debug/IoCheckoutFc/io.json
BlakeFreer, line 1: going forward, you can ignore the debug projects. they're going to be deleted soon

BlakeFreer (commented):
The structure of each element exists in three separate spots, ex dataclass `database.Bus`, class `database_config.BusConfig` and function `file_parser.parse_bus()`

The structures should each be described exactly once and I believe it'll look most like the current dataclass mixed with the config class. You should not need to write a `parse_` function or associated `Config` class. The dataclass should define its structure and config, and a higher-level system should define the parser.

Ex. You could write a `@decorator` which autogenerates the parsing / validation method given a dataclass-esque structure.

```python
# this is a very rough idea. syntax may vary
@configitem
class Bus:
    name = StringType(min_length=3)
    messages = ListType(Message)

@configitem
class Message:
    name = StringType
    id = IntegerType(signed=false)
```

The  `@configitem` decorator would inspect the class and generate a custom parser + validator which would look something like

```python
def Message.Config(input: dict):
    self.name = StringType.parse(input.pop("name")) 
    self.id = IntType.parse(input.pop("id"))

def Bus.Config(input: dict):
    self.name = StringType.parse(input.pop("name"))
    self.messages = [Message.parse(msg) for msg in input.pop("messages")]
```

Each `Type.parse` method would internally validate itself. You would then write the high-level types (ListType, StringType, IntType) with their parsers, but then the config structure can be really cleanly expressed as something looking like a dataclass

scripts/cangen/cangen/file_parser.py
```diff
@@ -0,0 +1,131 @@
...
+    EnumSignal,
+    FloatingSignal,
+    Endian,
+)
+from database_config import BusConfig, NodeConfig, EnumTypesConfig
+
+
+def _parse_signal(signal_data: Dict[str, Any], enum_types: Dict[str, Enum]) -> Signal:
+    base_attrs = {
+        "name": signal_data["name"],
```
BlakeFreer, line 21: you should use `data.pop("key")` since it properly "consumes" the value. it also lets you check that `data` is empty at the end of your parser so that you can alert the user if they've provided unknown values

`.pop` also is one way to implement default values ([docs](https://docs.python.org/3/library/stdtypes.html#dict.pop))

scripts/cangen/cangen/file_parser.py
```diff
@@ -0,0 +1,131 @@
...
+        "start_bit": signal_data["start_bit"],
+        "length": signal_data["length"],
+        "endianness": Endian[signal_data["endianness"]],
+        "description": signal_data.get("description"),
+        "additional_receivers": signal_data.get("additional_receivers"),
+    }
+
+    signal_type = signal_data["signal_type"]
+    if signal_type == "integral":
```
BlakeFreer, line 30: this looks like a good use for `match-case` <https://docs.python.org/3/tutorial/controlflow.html#match-statements>

scripts/cangen/cangen/database_config.py
```diff
@@ -0,0 +1,204 @@
...
+        # Convert string to float if needed
+        # This is needed for YAML parsing interpreting numerical values as strings (e.g. 1e-06)
+        if isinstance(value, str):
+            try:
+                value = float(value)
+            except ValueError:
+                raise ValueError(f"Cannot convert string '{value}' to float")
+
+        if not isinstance(value, (int, float)):
+            raise ValueError(f"Expected number, got {type(value)}")
```
BlakeFreer, line 98:
```suggestion

```

Just go straight to `val = float(val)`

scripts/cangen/cangen/database.py
```diff
@@ -1,76 +1,156 @@
...
+
+@configitem
 @dataclass
 class FloatingSignal(Signal):
-    is_signed: bool
-    value_range: tuple[float, float]
-    scale: float
-    offset: float
-    unit: str
+    is_signed: bool = config(validator=BooleanType(), required=False, default=None)
```
BlakeFreer, line 150: remove this, all floats are inherently signed

scripts/cangen/cangen/database_config.py
```diff
@@ -0,0 +1,204 @@
...
+import re
+
+T = TypeVar("T")
+
+
+class ConfigType(Generic[T], ABC):
+    def __init__(
+        self, type_: Type[T], required: bool = True, default: Optional[T] = None
+    ):
+        self.type = type_
```
BlakeFreer, line 13: where do you use this? is it necessary?

scripts/cangen/cangen/file_parser.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _parse_message(
+    message_data: Dict[str, Any], nodes: Dict[str, Node], enum_types: Dict[str, Enum]
+) -> Message:
+    # Parse and create a Message object with its signals and node references.
+    signal_data_list = message_data.get("signals", [])
+    message_data_copy = {k: v for k, v in message_data.items() if k != "signals"}
+
+    message = Message(message_data_copy, context={"nodes": nodes.keys()})
```
BlakeFreer, line 44:
```suggestion
    signal_data_list = message_data.pop("signals")

    message = Message(message_data, context={"nodes": nodes.keys()})
```

two things here:
1. the `signals` field should be required on each message. if the user wants it empty, they can explicitly write `signals: []` in the yaml
2. Use `.pop` to remove the element from the dictionary. then you don't need to make a copy (just continue using `message_data`). In general, you should always `pop` the data to consume it instead of temporarily `get`-ting it

scripts/cangen/cangen/file_parser.py
```diff
@@ -0,0 +1,118 @@
...
+
+    message.signals = [_parse_signal(s, enum_types) for s in signal_data_list]
+    message.sender = nodes[message_data["sender"]]
```
BlakeFreer, line 47:
```suggestion
    message.sender = nodes[message_data.pop("sender")]
```

scripts/cangen/cangen/file_parser.py
BlakeFreer, line 1: Can you move the `_parse_CLASS` method behaviour into the CLASS constructor? Seems very possible.

If not, at least make them each a `@staticmethod` of the respective class

scripts/cangen/cangen/file_parser.py
```diff
@@ -0,0 +1,118 @@
...
+    message.receivers = [nodes[receiver] for receiver in message_data["receivers"]]
+
+    return message
+
+
+def parse_bus(data: Dict[str, Any]) -> Bus:
```
BlakeFreer, line 53: relating to the file comment, it seems like the `Bus()` constructor should be doing more. Why separate out `messages, nodes, enum_types` from `data` only to use them to set fields on `bus`? Seems like this belongs to the constructor.

BlakeFreer (changes requested):

BlakeFreer:
See this comment first. https://github.com/macformula/racecar/pull/396#discussion_r1976807982

scripts/cangen/cangen/database.py
```diff
@@ -1,76 +1,156 @@
...
+    receivers: List[Node] = config(
+        validator=ListType(StringType(min_length=1)), default_factory=list
+    )
+    frequency: float = config(validator=FloatType(min_value=0.0))
+    description: Optional[str] = config(
+        validator=StringType(), required=False, default=None
+    )
 
+    @classmethod
+    def validate_config(cls, validated, context=None):
```
BlakeFreer, line 65: also validate:

1. signal names are unique
2. signal bits don't overlap
3. signal fits inside the message (given by `length`)

2 and 3 can be simultaneously validated by creating a list of strings with the length = number of __bits__ (not bytes) in the message (length * 8) and initializing all indices to `None`

For each signal, use `start_bit` and `bit_length` to set each index in the list to the signal's name. If the index is out of range, then the signal doesn't fit in the message. if the list item is not `None` then there is overlap.

Ex 8 bit message
```python
bit_owner: list[str] = {None} * 8 * message.length
for signal in signals:
    for i in range(signal.start_bit, signal.start_bit + signal.bit_length):
        try:
            if bit_owner[i] is not None:
                raise ValueError("signal {signal.name} and {bit_owner[i]} overlap at bit {i}")
            bit_owner[i] = signal.name
        catch IndexError:
            raise ValueError("signal {signal.name} does not fit in the message")
# can discard bit_owner now
```

scripts/cangen/cangen/database.py
```diff
@@ -1,76 +1,156 @@
...
+    )
+    additional_receivers: Optional[List[Node]] = config(
+        validator=ListType(StringType()), required=False, default=None
+    )
+    signal_type: str = config(validator=StringType(), required=True)
+
+    @classmethod
+    def validate_config(cls, validated, context=None):
+        # Validate signal type
+        if validated["signal_type"] not in ["integral", "boolean", "enum", "floating"]:
```
BlakeFreer, line 104: this is caught by the dictionary lookup in `_parse_signal`

scripts/cangen/cangen/database.py
```diff
@@ -1,76 +1,156 @@
...
+    description: Optional[str] = config(
+        validator=StringType(), required=False, default=None
+    )
+    additional_receivers: Optional[List[Node]] = config(
+        validator=ListType(StringType()), required=False, default=None
+    )
+    signal_type: str = config(validator=StringType(), required=True)
+
+    @classmethod
+    def validate_config(cls, validated, context=None):
```
BlakeFreer, line 102: make sure the signal "fits" in the number of bits
ex make sure `length` >= number of enum items

PR #404, opened 2025-02-12
LV Bug Fixes

BlakeFreer:
Problem: LV wasn't receiving CAN messages.
Fix: Enabled CAN RX0 interrupt (required by the stm32 mcal CAN driver)

Problem: State machine wasn't transitioning.
Source: HAL_GetTick was returning a `uint32_t` above the `int` range, messing with the elapsed time calculation.
Fix: Record the first timestamp and subtract off all subsequent GetTick calls

### DIFF
--- a/firmware/projects/LVController/main.cc
+++ b/firmware/projects/LVController/main.cc
@@ -103,37 +103,27 @@ public:
                     bindings::imu_gps_en.SetHigh();
                 }
 
-                if (elapsed > 50) transition = PWRUP_ACCUMULATOR_ON;  // temp
-
-                // flag = false;
-                // if (elapsed > 50) {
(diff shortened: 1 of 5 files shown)


sokosam (commented):
Look good but I think the elapsed time variables a little hacky. Could GetTick return a u_int32 instead of a int ?

BlakeFreer:
> Look good but I think the elapsed time variables a little hacky. Could GetTick return a u_int32 instead of a int ?

I cleaned up the implementation a bit by moving the static variable inside `GetTick()` (also makes the change for the stm32-ev6 platform). I don't want to use uint32 at the high level since we often subtract values and uint could lead to overflow

PR #410, opened 2025-02-24
Add OpenOCD flashing

### LINKED ISSUE #409: Add flashing over OpenOCD
The [stlink](https://github.com/stlink-org/stlink) tools can sometimes be tricky to set up. OpenOCD is already used for debugging in racecar and allows easy flashing.

samparent97:
closes #409

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -50,3 +50,6 @@ debug-openocd:
 
 debug-gdb:
 	arm-none-eabi-gdb "$(BUILD_DIR)/main.elf" -ex "target extended-remote localhost:3333" -ex "load"
+
+openocd-flash: build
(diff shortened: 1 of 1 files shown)


firmware/Makefile
```diff
@@ -50,3 +50,6 @@ debug-openocd:
 
 debug-gdb:
 	arm-none-eabi-gdb "$(BUILD_DIR)/main.elf" -ex "target extended-remote localhost:3333" -ex "load"
+
+openocd-flash: build
```
BlakeFreer, line 54: is there any reason not to call this `flash` and hide the openocd implementation detail?

samparent97: there is also `st-flash`

BlakeFreer (approved):

PR #417, opened 2025-02-26
Add tests for Arrays, replace CircularQueue with `etl::circular_buffer`

BlakeFreer:


firmware/shared/util/test_arrays.cc
```diff
@@ -0,0 +1,38 @@
...
+    EXPECT_EQ(idx, 4);
+
+    float min = GetMinimum<float, 6>(data, &idx);
+    EXPECT_EQ(min, -10.f);
+    EXPECT_EQ(idx, 2);
+
+    float avg = GetAverage<float, 6>(data);
+    EXPECT_FLOAT_EQ(avg, 2);
+
+    ASSERT_EQ(1, 1);
```
TylerStAmour, line 21: Is there any need for this? I'm not super familiar with gtest

BlakeFreer: You're right, that should be removed

luaibash (approved):
LGTM, changes from circular queue to the library version are simple and intuitive.

PR #419, opened 2025-02-27
Build and Test Projects in PRs

### LINKED ISSUE #418: Build and Test Projects in PRs
Ensure projects are built and tested automatically in PRs using GitHub Action workflows

The strategy matrix will have to be updated as this functionality is expanded to other projects

Currently, we are only going to build and test `FrontController` for `cli`

The command is `make PROJECT=FrontController PLATFORM=cli`

CameronBeneteau:
Closes #418 

Verified build commands work on macOS (`Demo/Blink` and `FrontController` for `cli`)

@BlakeFreer Let me know if you want to add more projects to the workflow matrix. If we're running these in PRs it might make sense to have as much coverage as possible so checks are automatically passed before reviews and merges.


.github/workflows/build-test-projects.yml
```diff
@@ -0,0 +1,28 @@
+name: Build and Test Projects
+on:
```
BlakeFreer, line 2: Can you restrict this to only run when there are changes in the firmware directory? 

See https://github.com/macformula/racecar/blob/main/.github/workflows/pr-format-check.yml

.github/workflows/build-test-projects.yml
```diff
@@ -0,0 +1,28 @@
...
+      matrix:
+        project:
+          - FrontController
+        platform:
+          - cli
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Set up Python
```
BlakeFreer, line 18: Is it worth caching python? I saw it was a 3 minute setup so it might be worth it

.github/workflows/build-test-projects.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
```
BlakeFreer, line 21: Do we need higher minimum for cangen? See its pyproject file

BlakeFreer (changes requested):

BlakeFreer:
Looks good, did you test it with a failing test (ex ASSERT_EQ(1,2)) to make sure that the action fails?

BlakeFreer:
Since the build process is relatively quick, can you add LV and TMS to the matrix? I'm not as worried about the Demos since they never go in the car

.github/workflows/build-test-projects.yml
CameronBeneteau: Yes added 3.10 minimum

CameronBeneteau:
> Looks good, did you test it with a failing test (ex ASSERT_EQ(1,2)) to make sure that the action fails?

https://github.com/macformula/racecar/actions/runs/13572520468/job/37940846312#step:5:215

.github/workflows/build-test-projects.yml
CameronBeneteau: First run (no cache): https://github.com/macformula/racecar/actions/runs/13573271494/attempts/1
Second run (cache): https://github.com/macformula/racecar/actions/runs/13573271494

BlakeFreer: Nice improvement

firmware/test.txt
BlakeFreer, line 1: Delete this

BlakeFreer (changes requested):
Remove the test file and we can merge

.github/workflows/build-test-projects.yml
CameronBeneteau: Separated into `setup-python` and `build-test` so python installations in `build-test` guarantee cache hits

Without `setup-python` it's possible all `build-test` matrix jobs will install cangen from scratch when cache misses (adding ~3 mins)

It's unlikely we'll be invalidating cache often so we could remove `setup-python` to simplify this script but let me know what you think

firmware/test.txt
CameronBeneteau: Yep was just cleaning up a few other things

CameronBeneteau:
@BlakeFreer Can you double check [this comment](https://github.com/macformula/racecar/pull/419#discussion_r1974157609) before approving

.github/workflows/build-test-projects.yml
BlakeFreer: I don't think we need to simplify the script. We won't be touching it much in the future

BlakeFreer (approved):

samparent97:
🔥

PR #432, opened 2025-03-05
Tarun/405 simplify battery monitor

Nagatarun25:
I combined two state machines into one and I updated the test cases accordingly. I also Updated the Enum names and only used BmSts enums.


firmware/projects/FrontController/control-system/src/battery_monitor.cc
```diff
@@ -43,160 +33,151 @@ std::optional<BmSts> BatteryMonitor::TransitionStatus(const Input& input,
...
             return ContactorCMD{
                 .precharge = OPEN,
-                .hv_positive = OPEN,
-                .hv_negative = CLOSED,
+                .positive = OPEN,
+                .negative = CLOSED,
             };
 
-        case ControlStatus::CLOSE_PRECHARGE:
+        case STARTUP_HOLD_CLOSE_NEG:
```
BlakeFreer, line 130: Since each `_HOLD_` state has the same output as the previous state, we can use switch-case "fallthrough" to make this code clearer.

```c++
case STARTUP_CLOSE_NEG:
case STARTUP_HOLD_CLOSE_NEG:
    return ContactorCMD{
        .precharge = OPEN,
        .positive = OPEN,
        .negative = CLOSED,
    };
```

when the switch-case block matches `STARTUP_CLOSE_NEG`, it will continue executing the switch until it sees a `return` or `break` which is how this works. it's useful since we explicitly want the two states to have the same output

do the same for the other `HOLD` states as well as `OPEN_PRECHARGE` and `RUNNING`

firmware/projects/FrontController/control-system/src/battery_monitor.cc
```diff
@@ -43,160 +33,151 @@ std::optional<BmSts> BatteryMonitor::TransitionStatus(const Input& input,
...
+                .negative = CLOSED,
             };
+
+        case RUNNING:
+            return ContactorCMD{
+                .precharge = OPEN,
+                .positive = CLOSED,
+                .negative = CLOSED,
+            };
+        case BmSts::LOW_SOC:
```
BlakeFreer, line 178: we always need to return something. it could be dangerous to let LOW_SOC and ERR_RUNNING return empty objects since we don't know what command it would represent and that could put the battery in a dangerous state

you should open all 3 contactors in either error state

firmware/projects/FrontController/control-system/test/test_battery_monitor.cc
BlakeFreer, line 1: in `CycleToState`, you should `assert` the state of `bm` before each `if(desired_state ... ) return bm;`. It would be really annoying if a test was failing (or worse, passing incorrectly) because we didn't verify `CycleToState`

ie. do this before each return check
```c++
auto out = bm.Update(... ); // store output of update

assert(out.status == BmSts::STATE);
if(desired_status = BmSts::STATE) {
    return bm;
}
```

since CycleToState is not a GoogleTest test, we use the regular `assert` instead of gtest's `ASSERT_EQ`

BlakeFreer (changes requested):

BlakeFreer (approved):

PR #435, opened 2025-03-07
423 software reset for can flash

sokosam:


firmware/projects/LVController/platforms/cli/bindings.cc
```diff
@@ -119,4 +119,12 @@ int GetTick() {
 void DelayMS(uint32_t milliseconds) {
     std::this_thread::sleep_for(std::chrono::milliseconds(milliseconds));
 }
+
+void SoftwareReset() {
+    std::cout << "entering software test" << std::endl;
```
BlakeFreer, line 124:
```suggestion
    std::cout << "Simulating software reset" << std::endl;
```

BlakeFreer (commented):
also need the function defined on FrontController for `cli`

BlakeFreer:
Code looks good but we should fix your formatting (this is my fault for not documenting it)

1. uninstall the `clang-format` extension - we don't use it
2. open a `.cc` file and open the vscode command palette. Search `format document with` and set `clangd` as the default c++ formatter
3. Make sure `format on save` is enabled for c++ files in your VS Code user configuration `.json`
4. Open and re-save each of the files in your PR to fix the formatting. You can see the specific formatting errors in the [PR Action](https://github.com/macformula/racecar/actions/runs/13716574731/job/38362579028?pr=435). (don't fix the errors manually, they should be fixed when you save with the proper formatter)
5. Run `pre-commit install`. We have a pre-commit check that should fix any formatting errors before they make it to GitHub. (We had problems with pre-commit crashing in the pas but that has been [fixed](#413)

sokosam:
I tried running pre-commit install in the terminal and I kept getting command not found.

BlakeFreer:
You'll need to install precommit with pip.

BlakeFreer (approved):

PR #437, opened 2025-03-13
DBC and code on LV & FC for CAN software reset 

sokosam:


firmware/projects/veh.dbc
```diff
@@ -232,10 +235,10 @@ BA_DEF_  "BusType" STRING ;
 BA_DEF_  "MultiplexExtEnabled" ENUM  "No","Yes";
 BA_DEF_DEF_  "BusType" "CAN";
 BA_DEF_DEF_  "MultiplexExtEnabled" "No";
+VAL_ 109 ECU 0 "FrontController" 1 "LvController";
```
BlakeFreer, line 238: also need `TMS`

firmware/projects/veh.dbc
```diff
@@ -35,6 +35,9 @@ BS_:
 
 BU_: PC_SG BMS FC TMS CANmod LVC RPI
 
+BO_ 109 InitiateCanFlash: 1 RPI
+ SG_ ECU : 0|1@1+ (1,0) [0|1] "" FC,LVC
```
BlakeFreer, line 39: this will need to be bigger for tms. there's no harm making it an 8 bit value even if we only use a few bits

firmware/projects/FrontController/main.cc
```diff
@@ -192,6 +192,14 @@ int main(void) {
         bindings::debug_led.Set(state);
         state = !state;
 
+        auto msg = veh_can_bus.GetRxInitiateCanFlash();
+
+        // _ecu is a boolean value, our enum is defined as 0 :
+        // FrontController, 1 : LVController
+        // In this case _ecu must be 0
+        if (msg.has_value() && !msg->ECU()) {
```
BlakeFreer, line 200: the fact that you need a comment indicates that this isn't the best structure. instead of doing a boolean evaluation, be more explicit and compare against the ECU enum value. see `PROEJCT/generated/can/veh_messages.h`

same for LV

BlakeFreer (changes requested):

BlakeFreer (approved):

CameronBeneteau:
🔥

PR #441, opened 2025-04-01
Disable Raspi Power on LV

BlakeFreer:
The Pi enable pin is now electrically tied high to ensure the Pi always has power. This is required to keep the Pi on while it performs a CAN Flash of the LV firmware.

This commit changes the pin from a GPIO output to reset state (high impedance) to avoid shorting 3.3V to ground.


CameronBeneteau (commented):
🔥

PR #445, opened 2025-04-06
Update create cubemx project gitignore template

Tegh25:
Original line:
```sh
!.board_config.ioc
```

This will not allow git to track the board_config.ioc file. Instead, the gitignore should allow any file with the .ioc extension to be tracked.
```sh
!*.ioc
```


BlakeFreer (approved):
good catch

Issue #345, opened 2024-11-30
Guard STM peripheral files


BlakeFreer:
The stm mcal files rely on files which are generated per-project by CubeMX. Thus if a project does not use a peripheral (say ADC), then the `mcal/stm/periph/adc.hpp` file will have a ton of errors since the file does not exist.

We can use `#ifdef` guards to detect if the current project uses each peripheral and enable / disable the whole file accordingly. CubeMX generates an `stm32f7xx_hal.h` file which includes `...hal_conf.h` which has `#define` statements to indicate which peripherals are enabled.

For an example, see `mcal/stm32f767/periph/can.hpp` and `can.cc` in the `dev/can` branch. https://github.com/macformula/racecar/blob/main/firmware/mcal/stm32f767/periph/can.hpp

```c++
//// can.hpp ////
#include "stm32f7xx_hal.h"
#ifdef HAL_CAN_MODULE_ENABLED // this disables the code if #define HAL_CAN is commented out below


//// stm32f7xx_hal.h ////
#include "stm32f7xx_hal_conf.h"


//// stm32f7xx_hal_conf.h ////
/* #define HAL_CRYP_MODULE_ENABLED */
#define HAL_ADC_MODULE_ENABLED
#define HAL_CAN_MODULE_ENABLED
/* #define HAL_CEC_MODULE_ENABLED */
/* #define HAL_CRC_MODULE_ENABLED */
```

# Task

1. ~Create a branch off of `dev/can`~

2. Guard all of the other peripheral files
3. ~Make a PR. Select `dev/can` as the base branch~

PR #446, opened 2025-04-07
Compile stm32f4 platform with f7 mcal

### LINKED ISSUE #383: Is the `stm32f767` platform compatible with other `stm` chips?
Can we use the current `stm32f767` platform to compile for other `stm32f` microcontrollers? If not, what is stopping us?

This would be awesome to have since Dashboard uses a different but similar chip, so reusing the mcal means we wouldn't have to rewrite it. 

Since we use the stm HAL, most of our code should be portable. Is there anything non portable? A good first check would be searching "767" in the `firmware/mcal` folder. 

Further, is it possible to generate CubeMX code for two different boards and have them fit into the same PROJECT-PLATFORM?

> Prior to #333, the answer was NO since we hard-coded the `mcpu` compiler flag into our build system. That PR improved it to use the generated flags, opening up this possibility.

Tegh25:
Closes #383.

### Commits:
<details><summary>Create stm32f4 project and compile with f7 mcal</summary>
<p>

Created the stm32f4 platform for the Demo/Blink project and used the existing stm32f767 platform mcal to compile it. STM defines macros which specify which platform is being built. Identifying these macros in the preprocessor allows for the correct HAL to be included without needing to edit the mcal when compiling a different platform.
</p>
</details>


BlakeFreer (changes requested):
Looking really good so far. Two notes

1. Some peripherals are different for different families, so can you make an f4 platform for the BasicIO, AnalogInput, AnalogPWM, and CAN/Send + CAN/Receive demo projects? For example, I don't think f4 has a CAN3 peripheral, so you will need to use the `#ifdef` guard to disable some code in `mcal/.../can.cc`. You should get a compiler error on F4 before fixing this error.
2. `clangd` is struggling to figure out which macro is defined, even though everything compiles fine. Notice how neither of the `#include`s are enabled.
image
I think this because the macro isn't defined until it's included, but it isn't included unless it's defined. Does this happen for you? If so, I think we should add
    ```cmake
    add_compile_definitions(STM32F7)
    ```
    to the `mcal_conf.cmake` file of each project->platform folder so that clangd explicitly knows which platform to parse.

Other than those, this looks really good. It'll help a lot with the Dashboard which is the f469. 

Can you also rename the mcal folder, cmake library, and namespace from `stm32f767` to `stm32f`? If you're in vscode, CTRL+SHIFT+F will be your friend to replace over the whole `firmware/` folder.

If it's not too much more to ask, can you complete #345 in this PR? It's just adding a single `#ifdef HAL_<PERIPH>` to avoid some compiler warnings about unused peripherals.

BlakeFreer:
I also noticed the clang-format test is failing. You should `pre-commit install` to get our git hook which fixes the formatting on each commit.

Even better, if you install [clangd18](https://github.com/clangd/clangd/releases?q=18.1.3&expanded=true) and set up the [VSCode clangd extension](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) then you can set `clangd` as your default c++ formatter and enable "format on save" to regularly fix the formatting

Tegh25:
Sure, I can take care of #345 in this PR.

BlakeFreer:
I added instructions to set up autoformatting. If you do this, then open and save each unformatted file to fix it
https://macformula.github.io/racecar/firmware/dev-setup/vscode/

Tegh25:
Hi @BlakeFreer, there's no `stm32f767` platform for the `Demo/AnalogPWM` project, so I skipped making the F4 platform for it. Let me know if you want me to make the F4 platform anyways.
All the other requested changes should be fulfilled.

BlakeFreer:
> Hi @BlakeFreer, there's no `stm32f767` platform for the `Demo/AnalogPWM` project, so I skipped making the F4 platform for it. Let me know if you want me to make the F4 platform anyways.
> 
> All the other requested changes should be fulfilled.

No that's fine. We will make it later if we need it

BlakeFreer:
There were a few projects which were missed in the conversion https://github.com/macformula/racecar/pull/446/commits/a8bd0080e3365e77dfd6b28cd050cf6290821026

I also updated the docs https://github.com/macformula/racecar/pull/446/commits/ba0de6d92749e2d3e152c373a1579195e6b7766e

BlakeFreer (approved):

Issue #440, opened 2025-03-28
C++ CAN Generation fails for non-byte aligned signals


sokosam:
The CAN message parser does not seem to work for messages which are not byte-aligned.

This works (the boolean messages start at multiples of 8):
```
BO_ 259 DashboardIndicatorStatus: 4 Dash
 SG_ IMDLed              : 16|1@1+ (1,0) [0|1] "" FC
 SG_ BMSLed              : 24|1@1+ (1,0) [0|1] "" FC
```

This doesn't:
```
BO_ 259 DashboardIndicatorStatus: 4 Dash
 SG_ IMDLed             : 11|1@1+ (1,0) [0|1] "" FC
 SG_ BMSLed             : 12|1@1+ (1,0) [0|1] "" FC
```

We were not able to receive the bits for IMD and BMS when they are on the same byte, but it works when the bytes are separated. Tested for FrontController and Dashboard buttons.

Our `cangen` python program generates `Tx` transmit messages for the sender node and `Rx` receive messages for the receiving node. The `Tx` generated code works fine; we can see the message being send correctly using [Busmaster](https://macformula.github.io/racecar/tutorials/busmaster/).

Therefore the problem is exclusively in the `Rx` message class which is failing to unpack the binary.

# Task

1. Recreate this error. Try using the `Demo/CAN` projects on CLI and spoofing your own messages. Edit the `demo.dbc` to create messages like `DashboardIndicatorStatus` which have the bug. Do not proceed until you have replicated the error.

2. Fix the `Rx` portion of [scripts/cangen/cangen/templates/messages.hpp.jinja2](https://github.com/macformula/racecar/blob/main/scripts/cangen/cangen/templates/messages.hpp.jinja2) to properly unpack these signals. This template file is used to generate `PROJECT/generated/can/messages.hpp`, so while editing the template, you can run `cangen path/to/project-folder` to regenerate the C++.

3. You should verify that the issue is gone by returning to part 1.

4. (optional, much more involved) Add some tests to `cangen/tests` to verify that the generated code can unpack messages with varying internal layouts. This will require using pytest to generate, compile, and run some GoogleTests.

PR #449, opened 2025-04-09
Fix cangen boolean unpacking & add tests

BlakeFreer:
The cause of issue #440 was a right shift on a value that was already type-casted to boolean. This wipes out the value.

To unpack a single bit from a byte (say at position 4 so mask = 0x10), we would call
```c++
unpack_right_shift<bool>(value, 4, 0x10);

template <typename T>
inline T unpack_right_shift(uint8_t value, uint8_t shift, uint8_t mask) {
    return static_cast<T>(static_cast<T>(value & mask) >> shift);
}
```

Suppose the byte value is `0xff` (all bits set so the boolean should be true). Watch how this proceeds

```txt
unpack_right_shift<bool>(0xff, 4, 0x10);
--> static_cast<bool>(static_cast<bool>(0xff & 0x10) >> 4);
--> static_cast<bool>(static_cast<bool>(0x10) >> 4);
--> static_cast<bool>(true >> 4);
--> static_cast<bool>(1 >> 4);
--> static_cast<bool>(0);
--> false
```

Therefore any boolean signal which needed a shift would always return 0 since the inner `static_cast<bool>` and `>>4` were doubly-shifting the bit value out of range.

# In this PR

- Edits the message template to handle bools separately ([diff](https://github.com/macformula/racecar/pull/449/commits/f96f2b1679779144156ac0ef0006d78f0d6bc620#diff-fad28dd6bb22eb3b653faea42f9e5662d7b364c6fa8be88fb5dff0ff9b6483d3R68))
- Removes the `unpack_left/right_shift()` functions altogether. It is so easy to generate their logic that we don't need to abstract them ([diff](https://github.com/macformula/racecar/pull/449/files#diff-c25efa4d34be225fb40371b5c9f556a2dbba0fe0d4cf67aa7ba52a3cc21fd9cd)).
- Only applies the scale and offset if they are non trivial (trivial being `value x 1 + 0`) which saves a few operations for most signals ([diff](https://github.com/macformula/racecar/pull/449/commits/f96f2b1679779144156ac0ef0006d78f0d6bc620#diff-fad28dd6bb22eb3b653faea42f9e5662d7b364c6fa8be88fb5dff0ff9b6483d3R79))
- Simplifies some of the Jinja template by inlining variables ([diff](https://github.com/macformula/racecar/pull/449/commits/f96f2b1679779144156ac0ef0006d78f0d6bc620#diff-fad28dd6bb22eb3b653faea42f9e5662d7b364c6fa8be88fb5dff0ff9b6483d3R73))
- Creates `Demo/TestCangen` which creates GoogleTests to ensure our generated code correctly unpacks CAN messages ([diff](https://github.com/macformula/racecar/pull/449/files#diff-9cc4aa60ca86fd2e8c64218bfc07ee6edb7bcf43d9b12c63dd0d9f4eb8636863))
    - You can run these tests with `make PROJECT=Demo/TestCangen PLATFORM=cli`


firmware/shared/comms/can/msg.hpp
```diff
@@ -23,17 +23,6 @@ concept TxMessage = requires(const T msg) {
...
 };
 
-// Used in the generated code to unpack from / pack into a RawMessage
-template <typename T>
-inline T unpack_right_shift(uint8_t value, uint8_t shift, uint8_t mask) {
-    return static_cast<T>(static_cast<T>(value & mask) >> shift);
-}
-
-template <typename T>
-inline T unpack_left_shift(uint8_t value, uint8_t shift, uint8_t mask) {
```
sokosam, line 33: I see the issue with the unpack functions but I'm wondering why there needs to be two static casts?

Correct me if I'm wrong but with your example of 0xff and 4, 0x10 
with the return being: 
```return static_cast<T>( (value & mask) >> 4);```
wouldn't we get the correct output?
```
unpack_right_shift<bool>(0xff, 4, 0x10);
--> static_cast<bool>((0xff & 0x10) >> 4);
--> static_cast<bool>((0x10) >> 4);
--> static_cast<bool>((0x01);
--> static_cast<bool>(0x01);
--> static_cast<bool>(1);
--> true
```

File: scripts/cangen/cangen/templates/messages.hpp.jinja2 [hunk 2]
```diff
@@ -52,42 +52,36 @@ private:
...
-        {% set sig_var_scale = "k" + sig.name + "Scale" %}
-        {% set sig_var_offset = "k" + sig.name + "Offset" %}
         {% for mask in masks %}
             {% set i = loop.index0 %}
             {% set shift = shifts[i] %}
             {% if mask != 0 %}
-                {% if shift >= 0 %}
-        {{ temp_sig_var }} |= shared::can::unpack_left_shift<{{ temp_sig_var_type }}>(raw_msg.data[{{ i }}], {{ shift }}U, {{ mask | decimal_to_hex }}U);
+                {% if T == "bool" %}
+        temp_{{ sig.name | camel_to_snake }} = raw_msg.data[{{ i }}] & {{mask | decimal_to_hex}};
```
sokosam, line 69: I'm guessing temp_x  would be recasted to bool later, LGTM.

scripts/cangen/cangen/templates/messages.hpp.jinja2
```diff
@@ -52,42 +52,36 @@ private:
...
                 {% else %}
-        {{ temp_sig_var }} |= shared::can::unpack_right_shift<{{ temp_sig_var_type }}>(raw_msg.data[{{ i }}], {{ -shift }}U, {{ mask | decimal_to_hex }}U);
+                    {% if shift >= 0 %}
+        temp_{{ sig.name | camel_to_snake }} |= static_cast<{{T}}>(raw_msg.data[{{ i }}] & {{mask | decimal_to_hex}}) << {{ shift }};
+                    {% else %}
+        temp_{{ sig.name | camel_to_snake }} |= static_cast<{{T}}>(raw_msg.data[{{ i }}] & {{ mask | decimal_to_hex }}) >> {{ -shift }};
+                    {% endif %}
```
sokosam, line 75: Isn't this part kind of the same thing I mentioned in msg.hpp ? Where its just a single static cast after shift?

sokosam (commented):
I think it looks fine, just wondering if some changes are necessary.

firmware/shared/comms/can/msg.hpp
BlakeFreer: there don't need to be two casts. the generated code now only has one

scripts/cangen/cangen/templates/messages.hpp.jinja2
BlakeFreer: it gets implicitly cast to bool during `signal_ = temp_signal` since `signal_` is bool

scripts/cangen/cangen/templates/messages.hpp.jinja2
BlakeFreer: Yes, we only need one shift so that's all i included

firmware/shared/comms/can/msg.hpp
BlakeFreer: This was the wrong answer. Your suggestion works in all most cases except 64 bit numbers, since it is possible to overflow and lose information with a left shift operation

PR #450, opened 2025-04-10
Add Hash version for DBC files to sync for avoiding CAN bus inconsistencies

luaibash:
## Description
- The purpose of this PR is to add hash version generation for dbc generated files such that these are sent to boards to compare and ensure that the boards are using the same dbc file, guarenteeing there will be no errors.
- Added `_extract_dbc_hash_version` in `can_generator.py` to create a hash version for message files, based on the raw contents of the file. If anything changes, the version will change.
- Added addition of hash version to message files, creating a constant that is set to the value of `_extract_dbc_hash_version`.
- Added a new bus message that is sent by LV and received by FC to send and compare hash versions
- Modified LVController to send the hash version, and the FrontController to receive the hash version, and compare it with it's own hash version to ensure that it is the same. If it is not, it will throw an error.

### DIFF
--- a/firmware/projects/FrontController/main.cc
+++ b/firmware/projects/FrontController/main.cc
@@ -63,6 +63,8 @@ VehicleDynamics vd{pedal_to_torque};
 // Global Governer input defined
 Governor::Input gov_in{};
 
+TxFC_Status fc_status{0,0,0,0,0,static_cast<uint8_t>(HashStatus::WAITING)};
+
 void UpdateControls() {
     // NOTE #1: For defining inputs, I commented out inputs where I didn't know
     // what to set it to NOTE #2: Some binding values like brake_pedal_front and
@@ -81,13 +83,11 @@ void UpdateControls() {
     // do I set some value?
     Governor::Output gov_out = gov.Update(gov_in, time_ms);
 
-    veh_can_bus.Send(TxFC_Status{
-        .gov_status = static_cast<uint8_t>(gov_out.gov_sts),
-        .di_status = static_cast<uint8_t>(gov_in.di_sts),
-        .mi_status = static_cast<uint8_t>(gov_in.mi_sts),
-        .bm_status = static_cast<uint8_t>(gov_in.bm_sts),
-        .user_flag = false,  // bindings::start_button.Read(),
-    });
+    fc_status.gov_status = static_cast<uint8_t>(gov_out.gov_sts);
+    fc_status.di_status = static_cast<uint8_t>(gov_in.di_sts);
+    fc_status.mi_status = static_cast<uint8_t>(gov_in.mi_sts);
+    fc_status.bm_status = static_cast<uint8_t>(gov_in.bm_sts);
+    fc_status.user_flag = false;  // bindings::start_button.Read(),
 
     // Driver Interface update
     DriverInterface::Input di_in = {
@@ -184,12 +184,22 @@ int main(void) {
 
     auto hash_version = veh_can_bus.GetRxSyncHashVersion();
     while (!hash_version.has_value()) {
+        fc_status.hash_status = static_cast<uint8_t>(HashStatus::WAITING);
+        veh_can_bus.Send(fc_status);
+
+        bindings::DelayMs(100);
+
         hash_version = veh_can_bus.GetRxSyncHashVersion();
     }
     if (hash_version->HashVersion() != generated::can::kVehDbcHashVersion) {
-        // Handle error
-        return -1;
+        while (true) {
+            fc_status.hash_status = static_cast<uint8_t>(HashStatus::INVALID);
+            veh_can_bus.Send(fc_status);
+
+            bindings::DelayMs(100);
+        }
     }
+    fc_status.hash_status = static_cast<uint8_t>(HashStatus::VALID);
 
     bindings::dashboard_power_en.SetHigh();
 
@@ -197,6 +207,7 @@ int main(void) {
 
     while (true) {
         UpdateControls();
+        veh_can_bus.Send(fc_status);
 
         bindings::debug_led.Set(state);
         state = !state;
--- a/firmware/projects/FrontController/main.cc
+++ b/firmware/projects/FrontController/main.cc
@@ -63,7 +63,8 @@ VehicleDynamics vd{pedal_to_torque};
 // Global Governer input defined
 Governor::Input gov_in{};
 
-TxFC_Status fc_status{0, 0, 0, 0, 0, static_cast<uint8_t>(HashStatus::WAITING)};
+using HashStatus = TxFC_Status::HashStatus_t;
+TxFC_Status fc_status{0, 0, 0, 0, HashStatus::WAITING};
 
 void UpdateControls() {
     // NOTE #1: For defining inputs, I commented out inputs where I didn't know
@@ -87,7 +88,6 @@ void UpdateControls() {
     fc_status.di_status = static_cast<uint8_t>(gov_in.di_sts);
     fc_status.mi_status = static_cast<uint8_t>(gov_in.mi_sts);
     fc_status.bm_status = static_cast<uint8_t>(gov_in.bm_sts);
-    fc_status.user_flag = false;  // bindings::start_button.Read(),
 
     // Driver Interface update
     DriverInterface::Input di_in = {
@@ -184,7 +184,7 @@ int main(void) {
 
     auto hash_version = veh_can_bus.GetRxSyncHashVersion();
     while (!hash_version.has_value()) {
-        fc_status.hash_status = static_cast<uint8_t>(HashStatus::WAITING);
+        fc_status.hash_status = HashStatus::WAITING;
         veh_can_bus.Send(fc_status);
 
         bindings::DelayMs(100);
@@ -193,13 +193,13 @@ int main(void) {
     }
     if (hash_version->HashVersion() != generated::can::kVehDbcHashVersion) {
         while (true) {
-            fc_status.hash_status = static_cast<uint8_t>(HashStatus::INVALID);
+            fc_status.hash_status = HashStatus::INVALID;
             veh_can_bus.Send(fc_status);
 
             bindings::DelayMs(100);
         }
     }
-    fc_status.hash_status = static_cast<uint8_t>(HashStatus::VALID);
+    fc_status.hash_status = HashStatus::VALID;
 
     bindings::dashboard_power_en.SetHigh();
 
--- a/firmware/projects/FrontController/main.cc
+++ b/firmware/projects/FrontController/main.cc
@@ -63,8 +63,8 @@ VehicleDynamics vd{pedal_to_torque};
 // Global Governer input defined
 Governor::Input gov_in{};
 
-using HashStatus = TxFC_Status::HashStatus_t;
-TxFC_Status fc_status{0, 0, 0, 0, HashStatus::WAITING};
+using DbcHashStatus = TxFC_Status::DbcHashStatus_t;
+TxFC_Status fc_status{0, 0, 0, 0, DbcHashStatus::WAITING};
 
 void UpdateControls() {
     // NOTE #1: For defining inputs, I commented out inputs where I didn't know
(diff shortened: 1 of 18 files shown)


firmware/projects/veh.dbc
```diff
@@ -45,6 +45,9 @@ BO_ 3221225472 VECTOR__INDEPENDENT_SIG_MSG: 0 Vector__XXX
  SG_ FC_userBtnL : 0|1@1+ (1,0) [0|1] "" Vector__XXX
  SG_ FC_userBtnR : 0|1@1+ (1,0) [0|1] "" Vector__XXX
 
+BO_ 123 SyncHashVersion: 8 LVC
```
luaibash, line 48: Is this message written correctly? What do I put for the ID, I just put `123` for now.

BlakeFreer: If it compiles then it's written fine. As long as 123 isn't used elsewhere in this DBC the you're good to claim it

scripts/cangen/cangen/templates/messages.hpp.jinja2
```diff
@@ -18,7 +18,11 @@ namespace generated::can {
 {% set bus = bus_name+"Bus" %}
 class {{ bus }};
 
-/* -------------------- ENUMS -------------------- */
+/* --------------------- CONSTANTS --------------------- */
+
+const uint64_t {{ bus_name|upper }}_DBC_HASH_VERSION = {{ hash_version }}U;
```
BlakeFreer, line 23: can you rename the constant according to https://google.github.io/styleguide/cppguide.html#Constant_Names?

firmware/projects/FrontController/main.cc
```diff
@@ -180,6 +180,13 @@ void UpdateControls() {
 }
 
 int main(void) {
+    auto hash_version = veh_can_bus.GetRxSyncHashVersion();
```
BlakeFreer, line 183: this needs to happen after bindings::initialize, else the CAN peripheral isn't set up to receive

firmware/projects/FrontController/main.cc
```diff
@@ -180,6 +180,13 @@ void UpdateControls() {
...
+    if (hash_version.has_value() &&
```
BlakeFreer, line 184: this check will not occur if LV doesn't send the message soon enough. You should instead loop until the msg has a value

firmware/projects/FrontController/main.cc
```diff
@@ -180,6 +180,13 @@ void UpdateControls() {
...
+        hash_version->HashVersion() != generated::can::VEH_DBC_HASH_VERSION) {
+        // Add error throwing here
```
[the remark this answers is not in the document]
BlakeFreer: we can't throw an error in firmware since the system has to continue executing _something_. You should add a 2-bit signal `HashStatus` to the `FC_Status` message for system check . set up an enum in the DBC:

```
// add this near bottom of veh.dbc
VAL_ 112 HashStatus 0 "WAITING" 1 "VALID" 2 "INVALID"
```

we currently only transmit the status inside `UpdateControls`. You should change this by defining a global instance of `TxFcState` and sending it in both the main loop and your `waiting` loop

firmware/projects/LVController/main.cc
```diff
@@ -315,6 +315,8 @@ void UpdateBrakeLight() {
 }
 
 int main(void) {
+    veh_can.Send(TxSyncHashVersion(generated::can::VEH_DBC_HASH_VERSION));
```
BlakeFreer, line 318: needs to go after bindings::init

BlakeFreer, line 318: you should extend the LV state machine. add a state after the FrontController power is enabled where you send this message, and exit the state once you read `FcState.HashStatus == VALID` (see earlier comment on FC main)

BlakeFreer (changes requested):
good start!

firmware/projects/FrontController/main.cc
luaibash: Moved all hash code after bindings::initialize.

firmware/projects/FrontController/main.cc
luaibash: Added loop until the message has been received. I feel like this can be written in many different ways, so let me know if you would like it written differently.

firmware/projects/FrontController/main.cc
```diff
@@ -180,15 +180,17 @@ void UpdateControls() {
 }
 
 int main(void) {
+    bindings::Initialize();
+
     auto hash_version = veh_can_bus.GetRxSyncHashVersion();
-    if (hash_version.has_value() &&
-        hash_version->HashVersion() != generated::can::VEH_DBC_HASH_VERSION) {
-        // Add error throwing here
+    while (!hash_version.has_value()) {
```
BlakeFreer, line 186: this will never change, need to repeated call `GetRx...` as well

scripts/cangen/cangen/templates/messages.hpp.jinja2
luaibash: Updated constant name to follow guidelines.

firmware/projects/FrontController/main.cc
luaibash: `GetRx...` is called inside that while loop, it keeps calling and then checking if that value is retrieved

BlakeFreer: sorry i missed that, I saw the call before the loop and assumed it was the only one

firmware/projects/FrontController/main.cc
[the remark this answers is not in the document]
luaibash: Added `HashStatus` to `FC_Status` and defined a global instance of `TxFcState` such that it can be sent in the hash validation code.

firmware/projects/LVController/main.cc
luaibash: Extended the LV state machine to include a `PWRUP_CHECK_HASH_STATUS` state, which sends the hash, retrieves the FC_Status which holds the hash_status, and then checks if the message was received and if hash_status is valid, transitioning to the next state. If this is not true, it will just loop again.

firmware/projects/FrontController/main.cc
```diff
@@ -81,13 +83,11 @@ void UpdateControls() {
...
-        .di_status = static_cast<uint8_t>(gov_in.di_sts),
-        .mi_status = static_cast<uint8_t>(gov_in.mi_sts),
-        .bm_status = static_cast<uint8_t>(gov_in.bm_sts),
-        .user_flag = false,  // bindings::start_button.Read(),
-    });
+    fc_status.gov_status = static_cast<uint8_t>(gov_out.gov_sts);
+    fc_status.di_status = static_cast<uint8_t>(gov_in.di_sts);
+    fc_status.mi_status = static_cast<uint8_t>(gov_in.mi_sts);
+    fc_status.bm_status = static_cast<uint8_t>(gov_in.bm_sts);
+    fc_status.user_flag = false;  // bindings::start_button.Read(),
```
BlakeFreer, line 90: can you remove the user flag signal from the dbc? we don't need it now that the start button is connected to the dashboard instead of frontcontroller

firmware/projects/FrontController/main.cc
```diff
@@ -182,12 +182,32 @@ void UpdateControls() {
...
+    while (!hash_version.has_value()) {
+        fc_status.hash_status = static_cast<uint8_t>(HashStatus::WAITING);
+        veh_can_bus.Send(fc_status);
+
+        bindings::DelayMs(100);
+
+        hash_version = veh_can_bus.GetRxSyncHashVersion();
+    }
+    if (hash_version->HashVersion() != generated::can::kVehDbcHashVersion) {
+        while (true) {
```
BlakeFreer, line 195: can you add a comment in here to explain that this is an error state and that we intend not to exit? The `while(true)` may be intimidating to someone who doesn't know about the hash checking

scripts/cangen/cangen/can_generator.py
```diff
@@ -164,6 +165,14 @@ def _parse_dbc_files(dbc_file: str) -> Database:
     return can_db
 
 
+def _extract_dbc_hash_version(dbc_file: str) -> int:
+    """Generates a 64 bit integer hash of the dbc file"""
+    with open(dbc_file, "r") as f:
+        dbc_file_string = f.read()
+        hash_hexadecimal = hashlib.sha256(dbc_file_string.encode()).hexdigest()[:16]
+        return int(hash_hexadecimal, 16)
```
BlakeFreer, line 173: is it necessary to convert the hash to string, then back to int? I feel like `.sha256` would return something that already is integer-like

scripts/cangen/cangen/can_generator.py
```diff
@@ -231,6 +240,7 @@ def _generate_code(bus: Bus, output_dir: str):
         "tx_msgs": tx_msgs,
         "bus_name": bus.bus_name,
         "node_name": bus.node,
+        "hash_version": _extract_dbc_hash_version(bus.dbc_file_path),
```
BlakeFreer, line 243: hash and version are kind of repetitive. I think a better name is just `dbc_hash` . Can you rename `HashVersion` to `DbcHash` throughout the code? It's also more self-explanatory since it describes what is being hashed

BlakeFreer (commented):
There are some compiler errors since merging in my branch, mostly regarding how enums are handled (you don't need `static_cast<uint>(enum_val)` anymore)

firmware/projects/FrontController/main.cc
luaibash: Removed usage of user flag in FrontController and in the dbc.

firmware/projects/FrontController/main.cc
luaibash: Added comment for the hash retrieval and the hash error state.

scripts/cangen/cangen/can_generator.py
luaibash: Renamed all instances of `HashVersion` and `hash_version` to `DbcHash` and `dbc_hash`.

scripts/cangen/cangen/can_generator.py
luaibash: `.sha256` returns a hash object, where you must use `.digest` to get the raw binary and then convert to an int using `int.from_bytes(),` or use `.hexdigest` to get a hexadecimal string, which I opted for since it's cleaner to convert an int to.

Overall I think it's necessary to do it this way, there's not much support from the library for immediately getting an int out of the hash generation.

BlakeFreer:
Tested this by implementing FC and LV on the linux platform

Valid DBC
```text
  vcan0  0FF   [6]  00 00 00 00 00 00       // FC HashStatus = Waiting
  vcan0  06F   [6]  12 FD 02 00 00 00       // LV State = 0x12=18=PWRUP_CHECK_HASH_STATUS
  vcan0  07B   [8]  36 61 7D F4 08 87 37 67 // LV Sends hash
  vcan0  06F   [6]  12 87 06 00 00 00       // LV State = 0x12=18=PWRUP_CHECK_HASH_STATUS
  vcan0  07B   [8]  36 61 7D F4 08 87 37 67 // LV Sends hash
  vcan0  0FF   [6]  00 00 00 00 00 01       // FC HashStatus = 0x01 = VALID
  vcan0  0FF   [6]  00 01 00 00 00 01       // FC DiStatus advanced to 0x01
  vcan0  06F   [6]  12 48 24 00 00 00
  vcan0  07B   [8]  36 61 7D F4 08 87 37 67
  vcan0  06F   [6]  03 CA 09 00 00 00       // LV State = 0x03 (passed Hash check)
```

Invalid DBC (changed the dbc for LV and regenerated the C++ so it had a different hash)
```text
  vcan0  0FF   [6]  00 00 00 00 00 00
  vcan0  0FF   [6]  00 00 00 00 00 00
  vcan0  0FF   [6]  00 00 00 00 00 00
  vcan0  0FF   [6]  00 00 00 00 00 00
  vcan0  0FF   [6]  00 00 00 00 00 00          // FC waiting for hash
  vcan0  06F   [6]  12 15 02 00 00 00          // LV in PWRUP_CHECK_HASH_STATUS
  vcan0  07B   [8]  E0 1D 73 07 3F 41 E5 38    // LV sends hash
  vcan0  0FF   [6]  00 00 00 00 00 02          // FC enters 0x2 INVALID state
  vcan0  0FF   [6]  00 00 00 00 00 02
  vcan0  0FF   [6]  00 00 00 00 00 02
  vcan0  0FF   [6]  00 00 00 00 00 02
  vcan0  0FF   [6]  00 00 00 00 00 02
```

scripts/cangen/cangen/templates/messages.hpp.jinja2
```diff
@@ -19,6 +19,10 @@ namespace generated::can {
 {% set bus = bus_name+"Bus" %}
 class {{ bus }};
 
+/* --------------------- CONSTANTS --------------------- */
+
+const uint64_t k{{ bus_name|capitalize }}DbcHash = {{ dbc_hash }}U;
```
BlakeFreer, line 24:
```suggestion
const uint64_t k{{ bus_name|capitalize }}DbcHash = {{ dbc_hash | hex }}U;
```
This will write the hash as a hex value which makes it easier to verify against the CAN messages

```c++
const uint64_t kVehDbcHash = 7437561782157009206U;
// becomes
const uint64_t kVehDbcHash = 0x67378708f47d6136U;
```

compare the second to the CAN log

```
vcan0  07B   [8]  36 61 7D F4 08 87 37 67
```

BlakeFreer (changes requested):

PR #491, opened 2025-06-15
FreeRTOS demo

BlakeFreer:
We had had a set of shared interfaces and mcal implementations of
FreeRTOS / CMSIS-OS objects, but this limited our ability to use the
RTOS since our wrapper ommited most functionality.

This commit reverts to a regular FreeRTOS implementation


BlakeFreer:
TMS fail will be fixed in next pr, just an unused header

PR #496, opened 2025-06-16
Live tuning raspberry

sokosam:


firmware/projects/Dashboard/inc/Tuning.hpp
BlakeFreer, line 1: i don't think we need a tuning menu, the driver should just be able to go

scripts/canflash/CanFlashApp.py
BlakeFreer, line 1: ui looks good

scripts/canflash/scripts/tune.sh
BlakeFreer, line 1: we shouldn't have a separate script. this should be handled entirely in Python.

https://cantools.readthedocs.io/en/latest/

```python
import can

can_bus = can.interface.Bus('vcan0', bustype='socketcan')
data = example_message.encode({'Temperature': 250.1, 'AverageRadius': 3.2, 'Enable': 1})
message = can.Message(arbitration_id=example_message.frame_id, data=data)
can_bus.send(message)
```

Since this is a realtime system, we should send the tuning message periodically instead of just once.

Here's a pattern you could try:

1. Start a thread which sends the Tuning Message every 500 ms.

2. From the UI, when the user clicks "Submit", change the message that the thread is sending.

3. Add a toggle switch to enable/disable sending the message

Issue #507, opened 2025-10-01
SLS095 Linear Potentiometer Module


BlakeFreer:
we have linear potentiometers which measure the displacement of our suspension. This issue will write a driver to convert Voltage inputs (in `volts`) to a displacement output (in `mm`). 

1. Create a new folder `lib/sls095`. Copy the `lib/lookup_table` structure.
    > - Your code should code in a file called `sls095.hpp`
    > - everything should be in the `macfe::sls095` namespace
2. Write a function which takes a voltage input and returns position. 
    > - The sensor model is `SLS095 / 0075 / 3K0 / R / 66`
    > - The datasheet is <https://www.cw-industrial.com/sites/default/files/home/industrial-sensors/linear-position/SLS095-Technical-Information.pdf> (include this link in a comment at the top of your C++ file)
3. Add a couple tests to the `lib/sls095` folder. Make sure they run and pass when you execute `pio test` from the folder
4. Update `lvcontroller/src/suspension` to use the new function. See that file's comment ` // todo: update the mapping`
    > To use the `lib/sls095` module in `projects/lvcontroller`, you should add a relative symlink inside `projects/lvcontroller/lib` to the module. Use these commands
    > ```bash
    > cd projects/lvcontroller/lib
    > ln -s ../../../lib/sls095
    > ```
    > Then you should be able to `#include "sls095.hpp"` from anywhere in the LV controller project

# Considerations

- avoid using unnamed constants in equations. use c++ constants to give names to constants
- your function signature should clearly express the physical units (for example, `ComputeTime` is a bad name whereas `ComputeSeconds` explicitly shows the units. this is just an example, your code uses displacement and voltage, not Time)

PR #512, opened 2025-10-06
added positionfromV function and implemented in LV Controller

ManushPatell:
I had a question regarding the supply voltage. I couldn't see from the data sheet that there should a set voltage the linear potentiometer should operate at. I've kept it as a two parameter function (supplyVoltage and inputVoltage)


projects/lvcontroller/src/suspension/suspension.cc
```diff
@@ -1,17 +1,24 @@
...
+    5.0f;  // set to 5V, change depending on what the sensor runs at
 
 static void Measure(void) {
-    // todo: update the mapping
-    travel3 = bindings::suspension_travel3.ReadVoltage();
-    travel4 = bindings::suspension_travel4.ReadVoltage();
+    v3 = bindings::suspension_travel3.ReadVoltage();
+    v4 = bindings::suspension_travel4.ReadVoltage();
+
+    // Map v3, v4 to position (mm)
```
BlakeFreer, line 19: Before you write any comment, ask yourself: "Could I rename or restructure this code to convey the same information as the comment?"

If you rename the function name to `VoltToMillimeter`, then you won't need this comment

lib/sls095/sls095.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <span>
+#include <type_traits>
+
+namespace macfe {
+
+inline constexpr float stroke_mm =
+    75.0f;  // Physical stroke length in millimeters
+
+float positionFromVoltage(float inputVoltage, float supplyVoltage) {
+    if (supplyVoltage <= 0) {
```
BlakeFreer, line 15: good safety check to avoid divide by zero!

lib/sls095/sls095.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#pragma once
+
+#include <span>
+#include <type_traits>
+
+namespace macfe {
+
+inline constexpr float stroke_mm =
+    75.0f;  // Physical stroke length in millimeters
```
BlakeFreer, line 12: is this comment needed? the variable name is fairly clear

projects/lvcontroller/src/suspension/suspension.cc
```diff
@@ -1,17 +1,24 @@
...
+#include "suspension.hpp"
 
 namespace suspension {
 
+static float v3 = 0;
+static float v4 = 0;
 static float travel3 = 0;
 static float travel4 = 0;
+inline constexpr float sensorSupply_V =
+    5.0f;  // set to 5V, change depending on what the sensor runs at
```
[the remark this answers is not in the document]
BlakeFreer: also, `inline constexpr` is more used for compile-time computations. this would be better as a `static const float SENSOR_SUPPLY_V`

it doesn't affect performance, but `const` is more standard for standalone values

lib/sls095/test/test_sls095.cc
```diff
@@ -0,0 +1,34 @@
+#include <gtest/gtest.h>
+
+#include "sls095.hpp"
+
+/*
+ https://www.cw-industrial.com/sites/default/files/home/industrial-sensors/linear-position/SLS095-Technical-Information.pdf
+*/
+
+TEST(VoltageToPosition, ZeroVoltage) {
```
BlakeFreer, line 9: Good tests!

lib/sls095/test/test_sls095.cc
```diff
@@ -0,0 +1,34 @@
+#include <gtest/gtest.h>
+
+#include "sls095.hpp"
+
+/*
+ https://www.cw-industrial.com/sites/default/files/home/industrial-sensors/linear-position/SLS095-Technical-Information.pdf
```
BlakeFreer, line 6: This datasheet link comment should be in the header file since that's where the SLS095 code is written.

BlakeFreer:
Since there's clang-formatting issues, I'm guessing you don't have `pre-commit` installed since it should handle all formatting before you push

see https://macformula.github.io/racecar/firmware/dev-setup/#pre-commit

ManushPatell:
I have clang-format extension downloaded from last year, do we need that now?

lib/sls095/sls095.hpp
```diff
@@ -0,0 +1,22 @@
+/// @author Manush Patel
+/// @date 2025-10
+
+/*
+ https://www.cw-industrial.com/sites/default/files/home/industrial-sensors/linear-position/SLS095-Technical-Information.pdf
+*/
+
+#pragma once
+
+namespace macfe {
```
BlakeFreer, line 10: sorry one more thing! this should be `macfe::sls905` (see point 1 of #507)

otherwise, it's not clear to the function caller what `VoltToMillimeter` pertains to

BlakeFreer (changes requested):

BlakeFreer (approved):
awesome! :moose:

PR #526, opened 2025-10-14
Lvgl9.3 update

furqaankhurram:
**Upgraded LVGL from 8.3 to 9.3**

**Updated platformio.ini to use lvgl@9.3**
Migrated all API changes throughout the codebase

**Fixed LVGL 9.3 API Compatibility**
Changed lv_img_create → lv_image_create
Changed lv_img_set_src → lv_image_set_src
Changed lv_obj_del → lv_obj_delete
Updated lv_disp_* → lv_display_* functions
Updated image descriptor format from lv_img_dsc_t to lv_image_dsc_t
Removed internal LVGL header includes, replaced with generic lvgl.h

**Fixed Logo Screen Issue**
Original RGB565A8 format incompatible with LVGL 9.3
Fixed Window Close Handling

**Created Setup and Run Scripts**

setup.sh - One-time setup for all dependencies
run_simulator.sh - Start FrontController in separate terminal
run_dashboard.sh - Start dashboard in separate terminal
run.sh - All-in-one script (optional)
Made all scripts executable with proper error handling

**Rewrote README.md**
Added beginner-friendly "Complete Beginner" section at top
Created visual two-terminal diagram

**Added exclude_helium.py script to disable ARM SIMD optimizations for Linux builds**
This pre-build script prevents compilation errors when building LVGL 9.3 on x86/Linux platforms.


projects/dashboard/src/main.cc
```diff
@@ -26,17 +28,43 @@ int main(void) {
...
+            std::cerr << "Exception in main loop: " << e.what() << std::endl;
+            break;
+        } catch (...) {
+            std::cerr << "Unknown exception in main loop" << std::endl;
+            break;
+        }
+#endif
     }
+
+    std::cout << "Shutting down dashboard..." << std::endl;
```
BlakeFreer: Unlike computers, microcontrollers always need to be running something. The `main` function is not allowed to `return

Our `main.cc` should also be platform independent. Having the `__linux__` guards starts to break this. I recognize that the `__linux__` code is not compiled with `platform=stm32`, but the point of having separate platforms is that they run the exact same `main()` code.

projects/dashboard/include/DriveModeMenu.hpp
```diff
@@ -2,6 +2,8 @@
 
 #include <src/misc/lv_area.h>
 
+#include <vector>
```
BlakeFreer: Embedded systems are memory constrained and _are not allowed to crash_.

C++ standard library containers like `std::vector` use dynamic memory allocation to adjust the size of the container. This could cause a fatal error in our system if the underlying `malloc` fails to find sufficient memory.

We use Embedded Template Library (etl) for statically-allocated replacements of standard library containers. See the [ETL's motivation](https://github.com/ETLCPP/etl?tab=readme-ov-file#motivation) for more explanation on why it's better than the standard library.

In particular, there is [`etl::vector`](https://www.etlcpp.com/vector.html) which you should use instead of `std::vector`

projects/dashboard/src/platforms/linux/bindings.cc
```diff
@@ -13,14 +14,22 @@
...
 }
 
 void hal_init(void);
 
+// Signal handler for Ctrl+C
+static volatile bool signal_received = false;
+void signal_handler(int signum) {
+    (void)signum;
+    signal_received = true;
+    std::cout << "\n🛑 Ctrl+C detected, shutting down...\n" << std::flush;
```
BlakeFreer: emojis aren't portable. we shouldn't use them in code or runner scripts

projects/dashboard/exclude_helium.py
```diff
@@ -0,0 +1,27 @@
...
+# Get the library source files
+lib_dir = os.path.join(env.subst("$PROJECT_LIBDEPS_DIR"), env.subst("$PIOENV"), "lvgl")
+
+# Filter function to exclude Helium and NEON files
+def skip_helium_neon(node):
+    filepath = node.get_path()
+    # Exclude any file in helium/ or neon/ directories
+    if '/helium/' in filepath or '/neon/' in filepath:
+        if filepath.endswith(('.S', '.s', '.c')):
+            print(f"⊗ Excluding: {os.path.basename(filepath)}")
```
BlakeFreer: remove the emojis

projects/dashboard/src/StatusBar.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // Clear the vector
+    warnings_.clear();
+
+    // Update visibility
+    UpdateVisibility();
+}
+
+bool StatusBar::IsVisible() const {
```
BlakeFreer: do we need these `...Visible...` methods? Should the status bar always be visible?

projects/dashboard/src/main.cc
BlakeFreer: `std::cout` isn't compatible with stm32. this line should be in the linux bindings::Shutdown

projects/dashboard/src/platforms/linux/bindings.cc
```diff
@@ -39,13 +48,20 @@ class KeyboardInput : public macfe::periph::DigitalInput {
                 keyStates_[event.key.keysym.sym] = true;
             } else if (event.type == SDL_KEYUP) {
                 keyStates_[event.key.keysym.sym] = false;
+            } else if (event.type == SDL_QUIT) {
+                quit_requested_ = true;
```
BlakeFreer: What is SDL_QUIT? Is is `q` or `esc` or something else? Can you add it to the instructions where you say `space` and `tab` are enter and scroll?

BlakeFreer (changes requested):
pretty cool! there's a few things to fix up and I had some questions. also it doesn't build on linux for me

projects/dashboard/src/platforms/linux/bindings.cc
furqaankhurram: `SDL_QUIT` is triggered when you click the X button on the window (not a keyboard key). I've added a "Keyboard Controls" section to the README that documents all input methods:

- Tab - Scroll
- Space - Select/Enter
- Ctrl+C - Quit (terminal)
- X button - Close window (SDL_QUIT event)

This should clarify the difference between keyboard controls and window controls.

projects/dashboard/src/Display.cpp
```diff
@@ -81,5 +82,4 @@ void Display::InnerChangeState(State new_state_) {
     }
 
     screen_->Create();
-    lv_obj_del(old_screen);
```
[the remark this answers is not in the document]
furqaankhurram: the `lv_obj_del(old_screen)` line isn't needed anymore because `Screen::Create()` now cleans and reuses the existing screen instead of creating a new one.

it's more efficient to clean the screen (remove its children) and reuse it rather than delete the old screen and create a new one each time we transition. This isn't a LVGL 9.3 requirement, just a design pattern that I tried to implement.

BlakeFreer (approved):
:moose:

PR #527, opened 2025-10-18
Convert TMS to use FreeRTOS

luaibash:
## Description
- Modified ioc file to enable FreeRTOS, copying the same configurations that FrontController uses
- Removed the CLI environment for TMS, only allowing the stm32f767 environment to be used
- Removed the CLI environment in `platformio.ini` and added libraries for FreeRTOS to be used, similar to how FrontController is implemented
- Added files that are needed by FreeRTOS such as `FreeRTOSConfig.h` and `freertos.c`
   - Removed functions in `stm32f7xx_it.h` that are now implemented by `FreeRTOSConfig.h`
- Modified `main.cc` to now use a 10 Hz FreeRTOS task, where the main function initializes this task and schedules it, and the task runs the update functions every 100 ms

## Testing
- Compiled locally with no errors
- TODO: Flash and test the TMS code on the vehicle


projects/tms/platformio.ini
luaibash: Copied from FrontController's ini file.

projects/tms/src/main.cc
```diff
@@ -11,6 +11,16 @@
 #include "periph/gpio.hpp"
 #include "temp_sensor/temp_sensor.hpp"
 
+// FreeRTOS
+#include "FreeRTOS.h"
+#include "task.h"
+
+static const size_t STACK_SIZE_WORDS = 2048 * 16;
```
luaibash: There is only one task and the work seems considerably smaller, does the stack size have to be this big? I've simply copied the same stack size that FrontController uses for now.

projects/tms/src/main.cc
```diff
@@ -11,6 +11,16 @@
...
+static const uint32_t PRIORITY_10HZ = 1;
```
luaibash: Set priority to 1 since it is the only task.

projects/tms/src/platforms/stm32f767/Src/freertos.c
luaibash: Copied from FrontController since this file is needed for the project to compile/run.

projects/tms/src/platforms/stm32f767/Inc/stm32f7xx_it.h
luaibash: These functions are removed since they are now implemented by `FreeRTOSConfig.h`

projects/tms/src/main.cc
BlakeFreer: this is fine. there's no reason to reduce static memory allocation until we start running out of memory. it's not like there's any other process running on the microcontroller

PR #529, opened 2025-10-24
setup of lvbms system

ManushPatell:
Just wanted to make the initial set up with Harsh on call. I'm going to grind through this after wednesday but he has time right now so I wanted to show him the functions on the datasheets and a template to work on.

### DIFF
--- a/projects/lvcontroller/src/lvbms/lvbms.cc
+++ b/projects/lvcontroller/src/lvbms/lvbms.cc
@@ -23,62 +23,6 @@ macfe::lv::LvBms::LvBms(macfe::periph::SpiMaster& spi) : spi_(spi) {}
  * function has been called
  **/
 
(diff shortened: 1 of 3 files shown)


BlakeFreer (approved):
🫎

ManushPatell:
I've written out the command codes that were on the 2950B datasheet. I know some codes have multiple versions where some include `ADBMS: 6380B Compatible`. 6830 seems to be a multicell battery stack monitor which we don't have so I think we can ignore those.

When writing the implementation of each command code, I strongly urge to recheck my hex conversion because there's a chance that 1/52 codes are wrong by 1 or more bits. Additionally, the codes `ADI1, AD2, ADV` are a little more complex. I've asked Sharvin to tackle this. @HarshPatel08, I think we can start to implement some of these functions now!

PR #536, opened 2025-11-01
fc suspension 

ManushPatell:
I based off my bms driver branch which was my mistake but once this is done, we can merge and look to merge the BMS driver set up.

### DIFF
--- a/projects/front_controller/src/main.cc
+++ b/projects/front_controller/src/main.cc
@@ -271,8 +271,6 @@ void task_10hz(void* argument) {
     (void)argument;
     static uint8_t tx_counter = 0;
 
-    suspension::task_10hz(veh_can_bus);
-
     TickType_t wake_time = xTaskGetTickCount();
 
     while (true) {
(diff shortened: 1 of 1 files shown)


BlakeFreer (changes requested):
this code isn't called anywhere. you need to run `task_10hz` in main.cc

projects/front_controller/src/main.cc
```diff
@@ -270,6 +271,8 @@ void task_10hz(void* argument) {
     (void)argument;
     static uint8_t tx_counter = 0;
 
+    suspension::task_10hz(veh_can_bus);
```
BlakeFreer: this is only called once. needs to go in the loop. follow what we did for lv

BlakeFreer (changes requested):
this code isn't called anywhere. you need to run `task_10hz` in main.cc

projects/front_controller/src/main.cc
ManushPatell: it looks like LV used `scheduler::run()` as a loop to call registered tasks, on FC it looks like FreeRTOS calls `task_10hz` using `xTaskCreateStatic()` and is delayed within the loop by `vTaskDelayUntil`. If this is correct, I'm not exactly sure how looking at LV would help. Aren't these different implementations?

PR #539, opened 2025-11-06
feat(mcal): implement SpiMaster peripheral for stm32f, lnx, and cli

priyaltaneja:
Add SpiMaster implementations across all three platforms:
- stm32f: HAL-based implementation with GPIO chip select control
- lnx: spidev interface with configurable mode and speed
- cli: interactive hex I/O for testing and simulation

Includes Init()/Setup() functions for future bindings integration.

### DIFF
--- a/lib/mcal/stm32f/spi.hpp
+++ b/lib/mcal/stm32f/spi.hpp
@@ -1,0 +1,117 @@
+/// @author Priyal Taneja
+/// @date 2025-10-19
+
+#pragma once
+
+#include <cstdint>
+
+#include "periph/spi.hpp"
+
+#ifdef STM32F7
+#include "stm32f7xx_hal.h"
+#elif defined(STM32F4)
+#include "stm32f4xx_hal.h"
+#endif
+
+#ifdef HAL_SPI_MODULE_ENABLED
+
+namespace mcal::stm32f {
+
+class SpiMaster : public macfe::periph::SpiMaster {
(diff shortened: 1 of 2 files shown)


lib/mcal/stm32f/spi.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// initialize the SPI peripheral
+    /// must be called before using any transmission methods
+    /// typically called from bindings::Initialize()
+    void Init() {
+        if (initialized_) {
+            return;
+        }
+
+        // initialize chip select pin as output, set high (inactive)
+        GPIO_InitTypeDef GPIO_InitStruct = {0};
```
BlakeFreer: glad you learned stm HAL pin initialization, but you don't need to configure the chip select pin manually. we will configure in in CubeMX so you can assume the Port/Pin is is initialized already

lib/mcal/stm32f/spi.hpp
```diff
@@ -0,0 +1,117 @@
...
+        ChipSelectHigh();
+    }
+
+    /// check if the spi peripheral has been initialized
+    bool IsInitialized() const {
+        return initialized_;
+    }
+
+    /// get the underlying hal spi handle
+    SPI_HandleTypeDef* GetHandle() const {
```
BlakeFreer: I don't think we need this. When would it be called?

PR #540, opened 2025-11-07
Transmit Git Hash over CAN

luaibash:
## Description
- Added python script to retrieve the git hash and unstaged changes status, writing them as variables to a cpp file.
- Configured each project to run the python script before running the project itself, such that the git values are available to be used by the project.
- Modified each project to send the git hash CAN message every 1ms, using the generated values from the cpp file.

### DIFF
--- a/scripts/build/generate_git_hash.py
+++ b/scripts/build/generate_git_hash.py
@@ -1,36 +1,31 @@
+Import("env")
+
 import subprocess
 from pathlib import Path
 
-# Repository root is two levels up from this script
-repo_root = Path.cwd().resolve().parents[1]
+include_path = Path(env.subst("$PROJECT_INCLUDE_DIR")) / "generated"
 
 def get_git_hash():
-    try:
-        # Executed command line command to retrieve a 7 character git hash
-        git_hash = subprocess.check_output(
-            ["git", "rev-parse", "--short=7", "HEAD"],
-            cwd=repo_root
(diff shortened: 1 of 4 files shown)


projects/veh.dbc
```diff
@@ -163,7 +163,7 @@ BO_ 410 LvGitHash: 5 LVC
  SG_ Commit : 0|32@1+ (1,0) [0|0] "" RPI
  SG_ Dirty : 32|1@1+ (1,0) [0|0] "" RPI
 
-BO_ 420 TmsGitHash: 5 LVC
+BO_ 420 TmsGitHash: 5 TMS
```
luaibash: Assuming this is a typo

scripts/build/generate_git_hash.py
```diff
@@ -0,0 +1,51 @@
+import subprocess
+from pathlib import Path
+
+# Repository root is two levels up from this script
+repo_root = Path.cwd().resolve().parents[1]
+
+def get_git_hash():
+    try:
```
luaibash: Let me know if we want to handle errors differently, through logging or anything, I've set it to just provide default git values for now.

BlakeFreer: you can remove the try-catch block and just run the subprocess. platformio will quit the build if the command fails. i think this is desirable behaviour

BlakeFreer (approved):
🫎

PR #561, opened 2026-03-25
Fix the dashboard

noahjaye:
- Fixed lvgl V9 by changing font memory allocation

- Fixed dashboard unresponsiveness by adding lgvl_tick_inc to SysTick (Note: cubemx deletes this on reconfiguration, something to fix in time)


projects/dashboard/src/platforms/linux/lv_conf.h
```diff
@@ -27,7 +27,7 @@
  *====================*/
 
 /** Color depth: 1 (I1), 8 (L8), 16 (RGB565), 24 (RGB888), 32 (XRGB8888) */
-#define LV_COLOR_DEPTH 16
+#define LV_COLOR_DEPTH 32
```
ManushPatell, line 30: any reason why we're increasing the color depth @noahjaye ?

projects/dashboard/src/platforms/stm32/lv_conf.h
```diff
@@ -27,7 +27,7 @@
  *====================*/
 
 /** Color depth: 1 (I1), 8 (L8), 16 (RGB565), 24 (RGB888), 32 (XRGB8888) */
-#define LV_COLOR_DEPTH 16
+#define LV_COLOR_DEPTH 32
```
ManushPatell, line 30: ^^

projects/dashboard/src/platforms/linux/lv_conf.h
noahjaye: No, it was not something that I intentionally changed

patem258:
Tested dashboard pin functionality, everything works fine and graphics looks good!

Issue #557, opened 2026-03-06
Data Acquisition on Raspberry Pi 


ManushPatell:
Implement a Raspberry Pi data logger that will record vehicle can data during testing and convert it into readable sensor values using the vehicle DBC 

## Requirements

The logger should:
- read CAN messages from the raspberry pi CAN interface (vcan0)
- Use candump (reads CAN messages from socket interface and prints them to terminal)
- decode messages using the veh DBC file ( use cantools)
- Log the decoded messages into a **CSV** file with timestamps
- store CSV logs locally on the Raspberry Pi

Target signals:
- Linear potentiometers
- IMU
- Wheel Speed sensors
- Temperature Sensors (Battery, Motor, Inverter)
- Battery Capacity

This shouldn't be too complicated code, all we want is to listen in on the socket interface, log important data, and store it in a csv file.

ManushPatell:
mentioned: #562

PR #562, opened 2026-03-25
feat: logger

HUANGV1:
Related to #557


ManushPatell:
.deb file is too large for github so for testing I can send that over. Additionally, I have the .env file with all the necessary info for set up... @noahjaye, @HUANGV1, @nicholasching

ManushPatell:
add bus voltage, bus current, and lv battery

PR #567, opened 2026-05-15
Update LV FW

ManushPatell:
need to change PH0, PH1 - HSE crystal oscillator


ManushPatell:
Need to remove dcdc dependancies

PR #569, opened 2026-05-18
FC Rev4 changes

noahjaye:


ManushPatell:
I made some HSD functions that will read the channel on my pr. We should use this on both fc and lvc. I think we can include it under lib and perhaps later put it in periph when we want to create platform and project independent versions of these. Right now I just care that we have it working for lvc and fc

ManushPatell:
Note: Need to implement HSE pin outs

PR #572, opened 2026-05-30
Dashboard rev4

noahjaye:


ManushPatell:
we can still use 0.1f, I'm not sure how to generate that though 
:()

projects/dashboard/src/platforms/stm32/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,2 @@
+eclipse.preferences.version=1
```
ManushPatell, line 1: remove this, not needed

projects/dashboard/include/DriveModeMenu.hpp
```diff
@@ -1,47 +1,140 @@
...
+    CANIndicator veh_can_;
+
+    FCStatusMessage fc_status_;
+    Speedometer speedo_;
+
+    uint32_t last_warning_time_{0};
     etl::vector<int, 10> active_warning_ids_;
-    int warning_cycle_index_;
-};
+    int warning_cycle_index_{0};
```
ManushPatell, line 139: is this brace initialization? I've never used it but if it is, is there a reason to opt for this and not "=0"

projects/dashboard/src/platforms/linux/lv_conf.h
```diff
@@ -420,7 +420,7 @@
 
 /** Add a custom handler when assert happens e.g. to restart MCU. */
 #define LV_ASSERT_HANDLER_INCLUDE <stdint.h>
-#define LV_ASSERT_HANDLER while(1);     /**< Halt by default */
+#define LV_ASSERT_HANDLER while(0);     /**< Halt by default */
```
ManushPatell, line 423: when lvgl hits an assertion error, it often means the system is in an unsafe state. The `while (0)` allows the program to run despite an assertion error and that's dangerous.

I'd rather have the system halt, and we could debug from there.

ManushPatell: Let's look into the idea of including a watchdog reset, I'll link this in an issue for someone to tackle

projects/dashboard/src/platforms/stm32/BSP/stm32469i_discovery_lcd.c
```diff
@@ -188,6 +189,22 @@ uint8_t BSP_LCD_Init(void) {
  *     - OTM8009A LCD Display IC Driver ititialization
  * @retval LCD state
  */
+static uint8_t led_counter = 0;
+
+static void advance_leds(void) {
+    ++led_counter;
+    HAL_GPIO_WritePin(LED1_GPIO_Port, LED1_Pin,
```
ManushPatell, line 196: Do we want to keep this advance leds function. This was for debugging so we can revert this or comment out and explicitly mention it's for debugging dash

projects/dashboard/src/platforms/stm32/bindings.cc
```diff
@@ -50,14 +71,15 @@ macfe::periph::DigitalInput& button_enter = mcal::button_enter;
 
 void Initialize() {
     HAL_Init();
+    uwTickPrio = 0;
+    HAL_NVIC_SetPriority(SysTick_IRQn, 0, 0);
```
ManushPatell, line 75: systick shouldnt be the most prioritized, revert the priorities of things back to normal. I don't think that was the solution to our dashboard bug

projects/dashboard/src/platforms/stm32/bindings.cc
```diff
@@ -66,35 +88,43 @@ void Initialize() {
...
-    // screen_driver_init();
+    advance_leds();
+    advance_leds();
+    advance_leds();
 }
 
 void DelayMS(uint32_t ms) {
-    HAL_Delay(ms);
+    // HAL_Delay(ms);
+    for (volatile uint32_t i = 0; i < 18000 * ms; i++);
```
ManushPatell, line 127: revert delay back

projects/dashboard/src/platforms/stm32/lv_conf.h
```diff
@@ -420,7 +420,7 @@
 
 /** Add a custom handler when assert happens e.g. to restart MCU. */
 #define LV_ASSERT_HANDLER_INCLUDE <stdint.h>
-#define LV_ASSERT_HANDLER while(1);     /**< Halt by default */
+#define LV_ASSERT_HANDLER while(0);     /**< Halt by default */
```
ManushPatell, line 423: revert on stm32 platform

projects/dashboard/src/main.cc
```diff
@@ -7,8 +7,8 @@
 #include "bindings.hpp"
 #include "generated/can/veh_bus.hpp"
 #include "generated/can/veh_messages.hpp"
+#include "gpio.h"
```
ManushPatell, line 10: remove unused gpio.h

projects/dashboard/src/StartingHV.cpp
```diff
@@ -6,7 +6,7 @@
 StartingHV::StartingHV(Display* display) : Screen(display) {}
 
 void StartingHV::CreateGUI() {
-    lv_obj_t* title_label = lv_label_create(frame_);
+    title_label = lv_label_create(frame_);
```
ManushPatell, line 9: any reason for this? don't see it changing anything, why remove it?

ManushPatell:
This is enough to get started, I don't want to spam :()

projects/dashboard/include/DriveModeMenu.hpp
noahjaye: No, I am happy to switch to = 0 if it improves readability

projects/dashboard/src/platforms/linux/lv_conf.h
noahjaye: That sounds good

projects/dashboard/src/platforms/stm32/BSP/stm32469i_discovery_lcd.c
noahjaye: Yeah good call, no point in having it normally.

projects/dashboard/src/platforms/stm32/bindings.cc
noahjaye: I think it was part of it, we'll test in the bay and go from there

projects/dashboard/src/platforms/stm32/bindings.cc
noahjaye: Yes

projects/dashboard/src/StartingHV.cpp
noahjaye: No reason now, but I think moving towards changing titles post-creation is a good thing to do

Issue #589, opened 2026-07-30
Create response for HVIL failure on LV Controller


ManushPatell:
### Problem

Currently the LV controller has no response when HVIL opens the accumulator contactors. After reading `READY_TO_DRIVE`, the LV state machine does not continue checking contactor feedback. Consequently, if HVIL breaks and the shutdown circuit opens the contactors, the LV controller will remain in `READY_TO_DRIVE` with the pumps, fans, DCDC (check if we need to have this state anymore), etc. 

The existing shutdown block is currently commented so `SHUTDOWN_DRIVER_WARNING, SHUTDOWN_PUMP_OFF, and SHUTDOWN_FAN_OFF` are currently unreachable. 

### Desired Behaviour
If theres any kind of fault that triggers contactors opening, we need to begin an appropiate shutdown sequence that turns the pumps off, fans off, and enters a defined safe state. 

### Requirements
- Monitor contactor feedback in `DCDC_ON` (wonder if we still need this, need to talk to LV), `POWERTRAIN_PUMP_ON, POWERTRAIN_FAN_ON and READY_TO_DRIVE`
- Enter `SHUTDOWN_PUMP_OFF` or a newly defined HV-loss state when the main contactors unexpectedly open
- This task can run at 100hz for now 
- After everything, we should switch to the state right before enable precharge state

This task is apart of a bigger response due to changes in FSAE rules. FSAE wants us to be able to re-precharge the pack without power cycling LV.

PR #598, opened 2026-09-03
HVIL Failure Response on LV

HarshPatel08:


ManushPatell:
addresses #589

projects/lvcontroller/src/main.cc
```diff
@@ -32,6 +32,7 @@ using LvState = TxLvStatus::LvState_t;
 
 static LvState state = LvState::PWRUP_START;
 static uint32_t elapsed = 0;
+static bool hv_fault_shutdown = false;  // true if shutdown was caused by contactors opening unexpectedly
```
ManushPatell: can you put this under fsm::Init()

projects/lvcontroller/src/main.cc
```diff
@@ -162,14 +163,34 @@ void Update_100hz(void) {
...
 
-            if (elapsed > 50) new_state = SHUTDOWN_COMPLETE;
+            if (elapsed > 50) {
+                if (hv_fault_shutdown) {
+                    hv_fault_shutdown = false;
+                    new_state = PWRUP_ACCUMULATOR_ON;
+                } 
+                
+                else {
+                    new_state = SHUTDOWN_COMPLETE;
```
ManushPatell: I'd argue no logic should go to shutdown_complete anymore. Our objective is if hvil breaks, theres a safe recovery to disable fans and pumps then going to a state prior to enable precharge. This state prevents us from having that.

projects/lvcontroller/src/main.cc
```diff
@@ -162,14 +163,34 @@ void Update_100hz(void) {
...
         case SHUTDOWN_COMPLETE:
             // there's no coming back from this
             break;
     }
 
+    bool check_shutdown = state == DCDC_ON || state == POWERTRAIN_PUMP_ON ||
+                      state == POWERTRAIN_FAN_ON || state == READY_TO_DRIVE;
+
+    if (check_shutdown && accumulator::ConfirmContactorsOpen()) {
+        hv_fault_shutdown = true;
```
ManushPatell: on second thought, why don't we avoid using hv_fault_shutdown, and just set the state to SHUTDOWN_PUMP_OFF. SHUTDOWN_DRIVER_WARNING doesn't do much, and keeps firmware waiting for an unnecessary amount of time. Let's remove shutdown driver warning as a state

ManushPatell:
we will test this soon 🦅

PR #599, opened 2026-09-04
sil architecture for tms

ManushPatell:


ManushPatell:
2 main things I wanted to bring up:

We have a collision with the macro `CRC`
- STM32 Hal defines a macro called `#define CRC    ((CRC_TypeDef *) CRC_BASE)`
- Cangen creates an accessor with the same name in `veh_messages.hpp`, inside of class RxCanFlashCRC called `inline uint8_t CRC() const { return crc_; }` 

I included a `#undef CRC` in app.cc, to include the HAL version of CRC. This will be a problem with other ECU's during re-architecture

Secondly, since we initially used `atomic_buffer.hpp` in `main.cc` and now we ported it over to `tms-common` logic, we ran into a problem with LDF not finding the header file atomic buffer because its now getting compiled into a library instead. I'm not too familiar with the concept, so I'm trying to learn more. Currently, the temporary fix is telling LDF that tms-common requires atomic buffer using a json file. I'm not personally a fan of this solution since it feels like we're just patching it and how this .json will have to be made for all ECU's re-architecture as well.

lib/tms-common/tms-common/tms-common.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "fan_controller.hpp"
+#include "generated/can/veh_bus.hpp"
+#include "periph/analog_input.hpp"
+#include "temp_sensor.hpp"
+
+namespace macfe::tms {
+
+using ::generated::can::TxBmsBroadcast;
+using ::generated::can::TxTMSValues;
+using ::generated::can::VehBus;
```
BlakeFreer: Don't use `using` in a header file. It pollutes the namespace. Write out the full path

It's ok to use in in the cc file

lib/tms-common/tms-common/tms-common.cc
```diff
@@ -0,0 +1,88 @@
...
+
+namespace macfe::tms {
+
+using namespace ::generated::can;
+
+TxTMSValues TMSBroadcast(const std::array<float, kSensorCount>& temperatures) {
+    return TxTMSValues{
+        .val1 = static_cast<uint8_t>(temperatures[0] * 50.0f),
+        .val2 = static_cast<uint8_t>(temperatures[1] * 50.0f),
+        .val3 = static_cast<uint8_t>(temperatures[2] * 50.0f),
```
BlakeFreer, line 14: What's the 50.0 factor here?

It should be a named constant

projects/tms/Inc/adc.h
```diff
@@ -1,51 +1,51 @@
-/* USER CODE BEGIN Header */
-/**
- ******************************************************************************
- * @file    adc.h
- * @brief   This file contains all the function prototypes for
```
BlakeFreer, line 5: Regenerate the code from cubemx and run post_cubemx.sh. These files shouldn't have a git diff. They should've just been renamed

projects/tms/Src/app.cc
```diff
@@ -0,0 +1,171 @@
...
+
+StaticTask_t t10hz_control_block;
+StackType_t t10hz_buffer[STACK_SIZE_WORDS];
+
+StaticTask_t t1hz_control_block;
+StackType_t t1hz_buffer[STACK_SIZE_WORDS];
+
+using namespace mcal::stm32f;
+
+AnalogInput temp_sensor_adc_1{&hadc1, SENS_1_UC_IN_CHANNEL};
```
BlakeFreer, line 46: Good, this is exactly how bindings should be replaced

lib/tms-common/tms-common/tms-common.cc
ManushPatell: This factor was already implemented prior. Additionally, the logic before made no sense. Take a look:

`TxTMSValues TMSBroadcast(const std::array<float, kSensorCount>& temperatures) {
    return TxTMSValues{
        .val1 = static_cast<uint8_t>(bindings::temp_sensor_adc_1.ReadVoltage() *
                                     50.0f),
        .val2 = static_cast<uint8_t>(bindings::temp_sensor_adc_2.ReadVoltage() *
                                     50.0f),
        .val3 = static_cast<uint8_t>(bindings::temp_sensor_adc_3.ReadVoltage() *
                                     50.0f),
        .val4 = static_cast<uint8_t>(bindings::temp_sensor_adc_4.ReadVoltage() *
                                     50.0f),
        .val5 = static_cast<uint8_t>(bindings::temp_sensor_adc_5.ReadVoltage() *
                                     50.0f),
        .val6 = static_cast<uint8_t>(bindings::temp_sensor_adc_6.ReadVoltage() *
                                     50.0f),
    };
}`

We pass in temperatures but read from bindings, which doesnt make sense. This version actually uses the passed in param

BlakeFreer: Yeah that is pretty bad. Global variables are not good

Glad to have it fixed. 

You should open a github issue "figure out what this TMS constant is"

BlakeFreer:
> Currently, the temporary fix is telling LDF that tms-common requires atomic buffer using a json file.

This is exactly what you should do. Each project needs to specify its dependencies. That's a good thing not "repetitive"

ManushPatell:
I autogenerated from stm32cubemx, and then ran the post_cubemx.sh script. I can notice this when I look at the files on github:
image

Otherwise I don't notice a difference...

BlakeFreer (approved):
nice

PR #601, opened 2026-09-08
move lv fault bits into LvAlerts. IMD and BMS faults are in LvAlerts …

### LINKED ISSUE #570: Porting Faults (IMD, BMS) to LvAlerts
I noticed this when creating a new fault/alert (overcurrent) for our HSD. Currently the new CAN message created runs at 10Hz as well, so I wanted to bring up if alerts should be sent at a higher rate or if they should be left as is. More importantly I think from a semantic point of view, I disagree with IMD, and BMS faults being stored as:  ``BO_ 211 LvStatus: 4 LVC``. Especially if they're being packed alongside info like `LvState, Counter, and MotorControllerState` I think moving forwards, we should port these over to `LvAlerts` instead...

HUANGV1:
…now as they are alerts rather than status. DAQ/docs/logger/front-controller mappings to match the new CAN messages layout.


ManushPatell:
🦅

PR #605, opened 2026-09-17
Noah/lv rearch 

noahjaye:
Continue rearchitecture, and finalize architectural decisions


ManushPatell:
@noahjaye  can you fix the branch issues so it's clearer what was done for lv rearch

lib/lv_common/lv_common/brakelight/brakelight.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace brake_light {
+struct brake_light_periph {
+    macfe::periph::DigitalOutput* brake_light_en;
+};
+class Controller {
+public:
+    void Init(macfe::periph::DigitalOutput& brake_light_en);
+
+    void task_100hz(generated::can::VehBus& veh_can);
```
ManushPatell, line 14: I think task specific code should remain in stm32 specific code so ideally our main.cc which should now be called app.cc. I know this doesn't force us to instantiate the task in this class

noahjaye: This is worth discussing, since we will likely want to test entire tasks in the SIL, which would mean duplicating task logic minus hardware-specific things.

noahjaye:
> @noahjaye can you fix the branch issues so it's clearer what was done for lv rearch

Did this, slightly gross and messes with times a bit but no more dashboard stuff

lib/lv_common/lv_common/brakelight/brakelight.hpp
ManushPatell: I agree with that stance, keep it

lib/lv_common/lv_common/accumulator/accumulator.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include "generated/can/veh_bus.hpp"
+#include "periph/gpio.hpp"
+namespace accumulator {
+
+struct accumulator_periph {
```
BlakeFreer, line 7: two things here:

- use references `DigitalOutput&`, not pointers `DigitalOutput*`. references cannot be null so you'll avoid a whole class of problems
- the accumulator periph struct is unnecessary. it just holds one item. just hold the DigitalOutput directly in the Controller class, no indirection

lib/lv_common/lv_common/brakelight/brakelight.cc
```diff
@@ -0,0 +1,20 @@
+#include "brakelight.hpp"
+
+#include "generated/can/veh_bus.hpp"
+#include "generated/can/veh_messages.hpp"
+#include "periph/gpio.hpp"
+
+namespace brake_light {
+using macfe::periph::DigitalOutput;
+
+void Controller::task_100hz(generated::can::VehBus& veh_can) {
```
BlakeFreer, line 10: don't create a class here. there's no data to store. prefer functions instead of classes when there is no state.

```c++
namespace brake_light {

void task_100hz(macfe::periph::DigitalOutput& brake_light_en, generated::can::VehBus& veh_can) {

}

}
```

In this comment https://github.com/macformula/racecar/pull/605/changes#r4072753415, I suggested flattening the namespace. If you don't have a class (as i'm suggesting here), definitely keep the namespace otherwise the function is just `macfe::lv::task_100hz` which is unacceptable and will certainly collide

in general, if you're just using a class to bundle functions under one name, use a namespace. I left a similar comment on `suspension`, though you should also use your judgement on what needs to be stateful (class) and what is just associated functionality (namespace)

lib/lv_common/lv_common/brakelight/brakelight.hpp
BlakeFreer: yeah it's fine to have fixed task rates in these classes

lib/lv_common/lv_common/accumulator/accumulator.hpp
```diff
@@ -0,0 +1,35 @@
...
+    macfe::periph::DigitalOutput* accumulator_en;
+};
+
+class Controller {
```
BlakeFreer, line 11: If everything is a controller, is anything a controller?

Firmware fundamentally just controls other devices. It'll get confusing later on when you're trying to remember "which `Controller` this is."

I'd rename these to the device they control. `accumulator::Controller -> accmulator::Accumulator`, `hsd::Controller -> hsd::Hsd`. There's no shame in duplicating the filename

Then you can flatten the namespace. These controllers should all be inside `namespace macfe::lv {}`

i.e. in app.cc you should see

```c++
using namespace macfe::lv; // this using is perfectly sensible in lv's source code

auto accumulator = Accumulator(accumulator_en);
```

whereas before you would have

```c++
auto accumulator_controller = accumulator::Controller(accumulator_en); // more things to write, worrying about a new namespace, uninformative `Controller` suffixes everywhere
```

lib/lv_common/lv_common/brakelight/brakelight.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "generated/can/veh_bus.hpp"
+#include "periph/gpio.hpp"
+
+namespace brake_light {
+struct brake_light_periph {
+    macfe::periph::DigitalOutput* brake_light_en;
+};
+class Controller {
```
BlakeFreer, line 10: see this comment https://github.com/macformula/racecar/pull/605/changes#r4072680642

get rid of this class. it stores nothing so might as well just be a function

lib/lv_common/lv_common/dcdc/dcdc.hpp
```diff
@@ -0,0 +1,39 @@
+/// @author Manush Patel
+/// @date 2026-05-24
+
+#pragma once
+
+#include "periph/analog_input.hpp"
+#include "periph/gpio.hpp"
+
+namespace dcdc {
+struct dcdc_periph {
```
BlakeFreer, line 10: as with this comment: https://github.com/macformula/racecar/pull/605/changes#r4072660308

i'd get rid of this struct and flatten it into the class arguments. you're not saving any typing by creating a struct.

i'd delete all of these `_periph` structs and directly store the peripherals in the class. It removes a layer of indirection. if you find a class that needs "too many" peripherals, then that class is probably doing too much and should be split apart. four peripherals is definitely not too many in this case

lib/lv_common/lv_common/fans/fans.cc
```diff
@@ -2,44 +2,45 @@
...
-#include "bindings.hpp"
 #include "etl/algorithm.h"
 #include "periph/gpio.hpp"
+#include "periph/pwm.hpp"
 
 namespace fans {
 
-static float power_setpoint = 0;
-static float current_power = 0;
+using macfe::periph::DigitalOutput;
```
BlakeFreer, line 11: do you even need these `using` statements? I don't see where you do. Remove them if not (and remove similar ones in other files)

lib/lv_common/lv_common/hsd/hsd.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace hsd {
+
+using macfe::periph::AnalogInput;
+using macfe::periph::DigitalOutput;
+
+struct hsd {
+    macfe::periph::AnalogInput* isense;
+    macfe::periph::DigitalOutput* isense_en;
+    macfe::periph::DigitalOutput** sel;
```
BlakeFreer, line 19: double pointer? that's scary.

these should all be references

lib/lv_common/lv_common/tssi/tssi.cc
```diff
@@ -0,0 +1,40 @@
...
+bool Controller::GetImdFault(void) {
+    return imd_fault;
+}
+
+bool Controller::GetBmsFault(void) {
+    return bms_fault;
+}
+
+/// @brief See FSAE 2025 Rules EV.5.11.5
+void Controller::task_10hz(int time_ms) {
```
BlakeFreer, line 18: you don't need time_ms if you know the task rate (10hz)

refactor this logic. instead of storing the absolute time you entered the state, reset a counter when you enter the state

```
state_entered_time = time_ms; --> time_in_state_ms = 0; 

(time_ms - state_entered_time) --> time_in_state_ms
```

then at the start / end of the function, add `time_in_state_ms += 100`

lib/lv_common/lv_common/tssi/tssi.hpp
```diff
@@ -0,0 +1,39 @@
...
+    macfe::periph::DigitalInput* bms_fault;
+};
+class Controller {
+public:
+    Controller(tssi_periph tssi_periph) : periph(tssi_periph) {};
+
+    void task_10hz(int time_ms);
+
+private:
+    void SetEnabled(bool enable);
```
BlakeFreer, line 22: these are private methods that just get/set a field.

delete them. an internal caller can just use the field, an external caller can't call the method since it's private.

do this for all other classes that have this error

lib/lv_common/lv_common/suspension/suspension.hpp
```diff
@@ -0,0 +1,23 @@
...
+        : periph(suspension_periph) {};
+    void task_10hz(generated::can::VehBus& veh_can);
+
+private:
+    suspension_periph periph;
+    const float SENSOR_SUPPLY_V = 3.3f;
+    float travel3 = 0.0f;
+    float travel4 = 0.0f;
+
+    void Measure(void) {}
```
BlakeFreer, line 20: there shouldn't be braces here. that makes it a function definition. just `void Measure(void);`

lib/lv_common/lv_common/scheduler/scheduler.cc
```diff
@@ -4,8 +4,6 @@
 
 #include <cstdint>
```
BlakeFreer, line 5: this is for a separate PR, but just mentioning it here

`scheduler` only exists because we wanted a non-freertos scheduler to let LV run on SIL.

Scheduler does nothing besides run tasks at some interval.

It should be replaced by FreeRTOS in the stm32 binary and custom logic for SIL. Then delete scheduler.hpp/.cc

Don't do this in this PR. That's a big separate issue. Make an issue for it an give it to a member once this re-arch goes in

lib/lv_common/lv_common/suspension/suspension.cc
```diff
@@ -1,24 +1,17 @@
...
 namespace suspension {
-
-static const float SENSOR_SUPPLY_V = 3.3f;
-static float travel3 = 0.0f;
-static float travel4 = 0.0f;
-
-static void Measure(void) {
-    float v3 = bindings::suspension_travel3.ReadVoltage();
-    float v4 = bindings::suspension_travel4.ReadVoltage();
+void Controller::Measure(void) {
```
BlakeFreer, line 7: as with `brake_light`, this class can also be collapsed to a method.

there's no reason to store travel3/4. it's not actually "state", it's just a holdover from `Measure` to `task_10hz`

This whole thing can be `void task_10hz(veh_bus, suspension_travel3, suspension_travel4)`

projects/lvcontroller/src/main.cc
BlakeFreer, line 275: where do you create all of the controller classes?

lib/lv_common/lv_common/motor_controller/motor_controller.hpp
```diff
@@ -0,0 +1,46 @@
...
+    void HandleSwitch(VehBus& veh_can);
+    void StateMachine_100hz(void);
+    motor_controller_periph periph;
+    bool enabled = false;
+
+    bool sw = false;
+    State state = State::OFF;
+    uint32_t elapsed = 0;
+};
+//! void Init(void);
```
BlakeFreer, line 44: delete this line

BlakeFreer (changes requested):

lib/lv_common/lv_common/dcdc/dcdc.hpp
noahjaye: I initially created the _periph structure for hsd, where we have 23(?) pins to keep track of, and then applied it to everything else for consistency. I could change hsd to create 6 classes with 3-4 inputs each, or just use the _periph struct for hsd.

lib/lv_common/lv_common/accumulator/accumulator.hpp
noahjaye: Point taken for pointers

lib/lv_common/lv_common/accumulator/accumulator.hpp
noahjaye: I wasn't sure if duplicating filename was encouraged but this is much cleaner

lib/lv_common/lv_common/brakelight/brakelight.cc
noahjaye: This makes sense and is definitely the right way to go

lib/lv_common/lv_common/fans/fans.cc
noahjaye: No, I believe not

lib/lv_common/lv_common/tssi/tssi.hpp
noahjaye: I am not sure why I did this

lib/lv_common/lv_common/scheduler/scheduler.cc
noahjaye: Understood, this clears things up significantly. This will be a good issue.

lib/lv_common/lv_common/suspension/suspension.cc
noahjaye: Agreed, keeping the namespace for consistency for all of these does not make sense, and collapsing one of either class or namespace makes sense

projects/lvcontroller/src/main.cc
noahjaye: I wanted to confirm that I had an idea of direction before moving on to the src/ directory, I will start that today.

lib/lv_common/lv_common/motor_controller/motor_controller.hpp
noahjaye: Yes

lib/lv_common/lv_common/dcdc/dcdc.hpp
BlakeFreer: Each High Side Driver is its own peripheral. I'd create a single class with the 3-4 peripheral each and create 6 instances of it. 

I don't know how the hsds are used, but it may not even make sense to have a "controller" object that holds the 6 hsds. They maybe should exist on their own. 

Ex if there is one HSD for fan power and one for pump cooling, then make your Fans and Pumps each take an HSD _as a peripheral_

```cpp
AnalogInput fan_hsd_isense{hadc1, ADC_CHANNEL1};
DigitalOutput fan_hsd_sel{GPIO_Port, GPIO_Pin};
HSD fan_hsd{fan_hsd_isense, fan_hsd_sel,...};
Fan fans{fan_hsd};
```

This is a good use of abstraction: multiple devices use HSDs, so define an HSD class once and reuse it inside each device

noahjaye: Agreed that this makes more sense. Looking into where the hsds are used, which determines where I put them

Issue #538, opened 2025-11-02
Raspberry Pi Provisioning


BlakeFreer:
Provide instructions for how to install the 64bit raspbian OS lite (non-desktop version) and set it up with the appropriate username (`macformula`), password, hostname (`racecar)`, and enable password-accessible SSH

After that is done, create a program (or use existing solutions like ansible) to automatically configure the entire pi. The program can be a single shell script, a python package, a Go binary, or anything else that can be easily `scp`'ed to the Pi and ran to set it up

Required features:

1. runs all updates (`sudo apt update / upgrade` should be one of the first commands)
2. configures `/boot/firmware/config.txt` for the CAN hat 

    ```
    dtparam=spi=on
    dtoverlay=mcp2515-can0,oscillator=16000000,interrupt=23
    dtoverlay=mcp2515-can1,oscillator=16000000,interrupt=25
    dtoverlay=spi-bcm2835-overlay
    ```
    
    > This file contains some important default. Instead of trying to edit / parse the file when provisioning, you should just replace the file with a complete configured copy which contains the defaults + your edits

3. creates a systemd service to automatically start our data acquisition program on bootup (talk to @TylerStAmour)
4. uses the wlan interface to host its own wifi network

    @CameronBeneteau set this up in 2024-25 but I forget what the process was. The raspberry pi would host a wifi network called `macfe-raspberry-pi` which you could connect your laptop to, then ssh into the Pi with `ssh macformula@racecar`. This solution is really nice since it doesn't rely on spotty MacWifi in the bay and can be used without any internet connection.

    We were also able to connect the raspberry pi to an additional wifi network to get internet.

    ```
    Internet
       |
       |(optional)
       |
       |
      RPi -------- Laptop
            (ssh)
    ```

PR #606, opened 2026-09-18
#538 - Raspberry Pi Provisioning

xavierruttendion:
Adds a script to provision the racecar Raspberry Pi with:
- Dual CAN interfaces
- DAQ startup on boot
- The `macfe-raspberry-pi` Wi-Fi hotspot

Setup instructions are included in README.


xavierruttendion:
Testing is still required

